PR #1166, opened 2015-01-13
Updated internal link under "share data"

cciciarelli8:
It's the open data policy link - not sure if this is the correct code for internal link

### DIFF
--- a/app/views/static/getinvolved.html.erb
+++ b/app/views/static/getinvolved.html.erb
@@ -170,7 +170,7 @@
           <article id="share-data" class="source-article">
             <header>
               <h2>Share data</h2>
-              <p>Global Forest Watch has an open data policy and commits to make our data, and that of our partners, open, discoverable, downloadable, and usable. We believe that open data enables the innovation, information sharing, analysis, and transparency required to support policies that better manage and sustain the world’s forests.</p>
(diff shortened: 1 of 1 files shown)


app/views/static/getinvolved.html.erb
```diff
@@ -170,7 +170,7 @@
           <article id="share-data" class="source-article">
             <header>
               <h2>Share data</h2>
-              <p>Global Forest Watch has an open data policy and commits to make our data, and that of our partners, open, discoverable, downloadable, and usable. We believe that open data enables the innovation, information sharing, analysis, and transparency required to support policies that better manage and sustain the world’s forests.</p>
+              <p>Global Forest Watch has an <a href="http://www.globalforestwatch.org/sources/data_policy" target="_blank">open data policy</a> and commits to make our data, and that of our partners, open, discoverable, downloadable, and usable. We believe that open data enables the innovation, information sharing, analysis, and transparency required to support policies that better manage and sustain the world’s forests.</p>
```
apercas: Hi! As this is internal is not needed to add the whole URL, as long as you start from the root: "/sources/data_policy" it will link to that page on both staging and production sites :)

mbarrenechea: Merged manually

PR #1167, opened 2015-01-13
Use this link to open data policy

cciciarelli8:


app/views/static/getinvolved.html.erb
```diff
@@ -170,7 +170,7 @@
           <article id="share-data" class="source-article">
             <header>
               <h2>Share data</h2>
-              <p>Global Forest Watch has an open data policy and commits to make our data, and that of our partners, open, discoverable, downloadable, and usable. We believe that open data enables the innovation, information sharing, analysis, and transparency required to support policies that better manage and sustain the world’s forests.</p>
+              <p>Global Forest Watch has an <a href="http://gfw-nav.herokuapp.com/sources/data_policy" target="_blank">open data policy</a> and commits to make our data, and that of our partners, open, discoverable, downloadable, and usable. We believe that open data enables the innovation, information sharing, analysis, and transparency required to support policies that better manage and sustain the world’s forests.</p>
```
apercas: Same here :) We only need the path after the domain (http://gfw-nav.herokuapp.com) starting with the slash. Thank you!

PR #1180, opened 2015-01-14
Updated data page text for the global mangroves

astrong19:
Updated the Date of Content to: 1997-2000, and updated the citation based on a request by UNEP-WCMC


alyssaw:
It looks like this PR was never merged with the redesigned site.  @apercas Can you take care of this?

apercas:
@alyssaw @astrong19 Fixed: Only the _sources file was updated, but not the _sources map one :)

PR #1215, opened 2015-01-20
How To: change link to feedback

beej23:


beej23:
@apercas can we actually leave the text change as is (leave the text about beta), and just change the link, as edited?

PR #1221, opened 2015-01-20
changes to stay informed page

alyssaw:


mbarrenechea:
Be careful with target="_blank. You missed close quotes: target="_blank"

PR #1224, opened 2015-01-21
How To: update accordion title

beej23:


beej23:
@apercas we also need to make sure these accordion titles don't mess up the formatting because they are too long. Can you check this?

PR #1256, opened 2015-01-23
updated links, button text, etc.

alyssaw:
@mbarrenechea I also added two new interesting helpers - did I do this right?


app/assets/javascripts/helpers/interestingHelper.js
```diff
@@ -31,61 +33,77 @@ define([], function() {
...
+      button_text: 'Contact us',
+    },
+    explore_applications: {
+      title: 'Explore GFW applications',
+      description: 'Explore GFW apps, tools designed to address specific forest issues.',
+      link: '/explore/applications',
+      button_text: 'Explore apps',
+    },
+      coming_soon: {
+      title: 'Coming soon to GFW,
```
apercas: This is no nice! But it lacks the end simple quote mark in this line :) We'll merge this after adding it!

PR #1343, opened 2015-02-17
Rankings: update 2012 with 2013 UMD

beej23:


apercas:
Added manually to develop.

PR #1373, opened 2015-03-04
Update LayerSpecModel.js

aagm:
just change layers order :D


apercas:
@aagm Hey, we're not working in feature/map-refactor anymore, could you add the PR to develop instead? Thank you!

PR #1398, opened 2015-03-09
MAP: minor text update (adding/removing space)

beej23:


beej23:
@apercas for some reason the title text in the summary section is not appearing bold. Can you add that?

PR #1423, opened 2015-03-12
Added the selector for QICC < 37

aagm:


apercas:
:dancers: :dancers: :dancers: :dancers:

PR #1441, opened 2015-03-18
Fix sharing snippets

adammulligan:
The sharing snippets for FB and G+ were not getting passed the supported browser check (the bot check in the user agent gem is just a regex match for "bot" :dizzy_face:) and the terms check, so I extracted that out and added some special cases for Facebook and g+. It's easy to add in more exemptions if we need to, for whatever reason. 

This seems like a big change, but I've just moved the user agent validation out in to a separate class as it's a separate responsibility from the app controller, and to make it testable.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,6 +1,6 @@
 class ApplicationController < ActionController::Base
 
-  helper_method :watch_cookie?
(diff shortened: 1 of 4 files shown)


apercas:
:dancer: :dancer: :dancer: :dancer:

PR #1453, opened 2015-03-23
Remove 1st carousel item

alyssaw:
Remove "GFW's 1-year anniversary" slide

@d4weed @apercas @mbarrenechea Let me know if I'm not doing this correctly!


d4weed:
It looks good,I thinkk we just need to add the "current" style to the new first item and it's good to go. @mbarrenechea can you look into this tomorrow?

PR #1455, opened 2015-03-24
Adds a download link mailer on mobile

adammulligan:
This PR introduces a download link mailer: when on a mobile device,
instead of linking directly to a Country download, we'll offer them the
ability to email themselves a link to the download.

The DownloadView on the client has been designed to be generic (it takes
any URL for a "download"), and so it can be used for mail notifications
for any type of download. The same goes for the mailer on the server.

screenshot 2015-03-24 15 38 52
screenshot 2015-03-24 15 31 30

### DIFF
--- a/app/assets/javascripts/countries/views/CountryShowView.js
+++ b/app/assets/javascripts/countries/views/CountryShowView.js
@@ -9,12 +9,13 @@ define([
   'mps',
   'scrollit',
   'views/SourceWindowView',
+  'views/DownloadView',
   'countries/views/CountryHeaderView',
   'countries/views/CountryShareView',
   'countries/models/CountryShowModel',
   'countries/helpers/CountryHelper',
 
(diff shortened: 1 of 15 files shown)


mbarrenechea:
Almost done! Styles should be more like this screenshot. You are using old styles. Change styles and I will seriously consider to accept your PR :)

screen shot 2015-03-24 at 17 25 00

apercas:
@adammulligan Also, could you split the PR into different commits? Getting everything in a single one can lead to some craziness when debugging :)

adammulligan:
@apercas sure thing. I actually squashed my WIP commits in to this one commit -- I normally do that because I find it makes the history easier to read, and I'm not sure people will find my small commit messages that useful :grin: 

I'll undo the rebase, though I'll have to open a new PR as I can't push to this branch after that :+1:

PR #1462, opened 2015-03-26
Send download emails on mobile

adammulligan:
Replaces the reverted #1456, check out that for a detailed description of the original feature.

This PR also unifies the small modal CSS so that it's shared nicely across the page, and is no longer terrible. I've also unified the Share dialog JS so that the map and country page use the same (hopefully nicer) code.

I forgot to mention in the other PR, this uses a mailer for the download emails and will need some sort of SMTP config, can someone (@apercas @mbarrenechea) with knowledge of the deploy process sort that out, or let me know how you normally configure stuff like that.


apercas:
@adammulligan Yes: you can check here gfw/config/application.rb that we're using Mandrill for everything email related.

adammulligan:
Oh, perfect, shouldn’t need any further config then.

apercas:
:dancers: :dancer: :dancer: :dancers: :dancer: :dancer: :dancers: :dancer: :dancer: :dancers: :dancer: :dancer: :dancers: :dancer: :dancer:

PR #1471, opened 2015-03-27
updated country names + order in several layers

alyssaw:


apercas:
This'll be deployed next Thursday.

alyssaw:
@apercas @mbarrenechea @d4weed This also needs to be deployed

PR #1472, opened 2015-03-30
Timeline fix for IE

adammulligan:
Very small fix for the map timeline width in IE 10/11. Parent element requires `width: auto`, and IE has some issues with that if the children don't explicitly define a width.


apercas:
:dancer: :dancer:

PR #1477, opened 2015-03-30
MAP: removing capital letters in data legend

beej23:
Please review this before merging!


apercas:
Hey, those files are automatically generated and they're stepped over once we deploy, so don't bother updating them :)

We have them just for back ups.

PR #1487, opened 2015-03-31
Map country layer toggle

adammulligan:
Adds toggles to the country layers in the map tab view. Quite a hacky solution, but with permission from @apercas :grinning: 

screenshot 2015-03-31 11 15 37


apercas:
:dancers: :dancers: :dancers:

PR #1495, opened 2015-03-31
Country layer toggle fix

adammulligan:
Uses present pubsub instead of guesswork :).


apercas:
Almost there :D

They should toggle as well when we click them from the drop-down menu, both sides. I know this could be tricky.

PR #1555, opened 2015-04-07
Map: toggle gain and loss and all forest change layers

mbarrenechea:


apercas:
:dancers: :dancers: :dancers: :dancers: :dancers:

PR #1570, opened 2015-04-08
Fit iframe to content of share previews

adammulligan:
On slower connections there can be a bit of a delay as we can't figure out the width until the iframe has loaded. Wasn't sure if it was worth adding some sort of loading indicator, thoughts? Here's how it looks: http://recordit.co/e3x3IqvTmY


mbarrenechea:
Sorry, we don't need this.

PR #1633, opened 2015-04-28
Map: other OSINFOR info window changes

beej23:


mbarrenechea:
It has been done manually.

PR #1718, opened 2015-06-09
Canada mining layers unified

aagm:


apercas:
:dancer: :maple_leaf:

PR #1807, opened 2015-08-26
updates to partnership page

alyssaw:
NOTE: logos need to be created for The Tilia Fund and Bobolink Foundation (we created ones before but I didn't know what to include for "class")


alyssaw:
@mbarrenechea @d4weed We need logos created for The Tilia Fund and the Bobolink Foundation, otherwise I think we're good to go!

PR #1887, opened 2015-10-22
Torque layer

adammulligan:
This PR introduces a Torque layer (with a matching Timeline). The timeline only roughly fits the design, this PR is mostly just to get the Torque layer reviewed early on. Since this is potentially going to replace other animated layers, it'd be nice if you could give the Torque timeline classes a thorough code review.

I've added a new layer to the spec for testing, which you can try out locally here: http://localhost:5000/map/3/0.54/20.80/ALL/grayscale/gfw_landsat_torque_alerts You'll have to reset your localStorage first, of course. The date shows as 1970 as the test table doesn't have real dates right now :(

Next up, date pickers!


apercas:
:dancer: :sparkles:

PR #1893, opened 2015-11-04
added new forest change layer: Guyra

aagm:


apercas:
For naming convention, could we change the name we're showing in the URL from 'guyra_deforestation' to just 'guyra'? thanks :)

PR #1901, opened 2015-11-10
Feature/plantation country layers

aagm:
note that there is a bug on peru, the wrapper interferes with the concession wrapper


aagm:
Hi i have check and there also another bug with this layer, as the infowindow appear correctly on the left menu but not on the legend

PR #1906, opened 2015-11-12
Fixes front-end tests

adammulligan:
I was getting some withdrawal symptoms (hands shaking, blurry vision, etc) after not writing any tests for a while, so I thought I'd get the test suite here up and running again. I didn't change any of the "real" code, just tests and config, so nothing should be broken.

* Updated the requirejs config for the tests to match the current front-end config
* Removed some spec files and test cases that no longer applied
* Updated a few test cases where the default values had changed in the code, but not the tests
* Fixed the problems with `window` variables required in some classes
* Removed jshint because it had, like, 1000 errors and I don't think any of us has time for that


apercas:
fucking superhero, @adammulligan :+1:

PR #1909, opened 2015-11-16
Adds FORMA 250 torque layer

adammulligan:
Adds a layer to test out the forma 250 data (for Peru). Mostly a standard layer addition, but also introduces a method to fit the map to the geometry bounds in the layerspec when selecting the layer from the menu.

@apercas This adds a new column to the layerspec: `fit_to_geom` that the MapPresenter uses to decide whether or not to try and fit to the geometry bounds. Does the staging server use `layerspec_nuclear_hazard`?

Test it out: http://localhost:5000/map/10/-5.75/-77.20/ALL/grayscale/forma_250?begin=2015-03-22&end=2015-11-16

### DIFF
--- a/app/assets/javascripts/map/cartocss/Forma250Layer.cartocss
+++ b/app/assets/javascripts/map/cartocss/Forma250Layer.cartocss
@@ -1,0 +1,11 @@
+#layer {
+  comp-op: lighter;
+  marker-fill-opacity: 1;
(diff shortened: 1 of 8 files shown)


adammulligan:
@apercas merging this now and deploying to staging so it can be shown off. Take a look at the code anyway when you're back and have a chance, thanks!

PR #1910, opened 2015-11-16
GLAD Torque CSS tweaks

adammulligan:


adammulligan:
@apercas I just pushed some changes to do this, if you could have a quick review when you have time? 🙏

apercas:
GLAD to see it moving forward :D

PR #1937, opened 2015-12-03
New layer fixes

adammulligan:
A few fixes for the forma/glad layers:

* start animation at the end so you can see the full extent of the changes first
* speed up the torque animations to match up with the existing loss/gain layer timing
* show the correct geographic coverage layer for forma 250

### DIFF
--- a/app/assets/javascripts/map/views/layers/Forma250CoverLayer.js
+++ b/app/assets/javascripts/map/views/layers/Forma250CoverLayer.js
@@ -1,0 +1,23 @@
+/**
+ * The Forma Coverage layer module for use on canvas.
+ *
+ * @return Forma250CoverLayer class (extends CanvasLayerClass)
+ */
+define([
+  'abstract/layer/CartoDBLayerClass',
+], function(CartoDBLayerClass) {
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/map/views/layers/Forma250CoverLayer.js
```diff
@@ -0,0 +1,23 @@
...
+define([
+  'abstract/layer/CartoDBLayerClass',
+], function(CartoDBLayerClass) {
+
+  'use strict';
+
+  var Forma250CoverLayer = CartoDBLayerClass.extend({
+
+    options: {
+      cartocss: '#layer { polygon-fill: #FF6699; polygon-opacity: 0.2; line-width: 0; }',
```
apercas: Even if it's just a single three attributes, can we have it in a different / separated file?

adammulligan: Yeah, good point! I just pushed a fix to this branch :+1:

apercas:
:tada: :)

PR #1945, opened 2015-12-10
Sources route has been replaced by data.gfw.org

adammulligan:
https://basecamp.com/3063126/projects/10728552/todos/219137373


adammulligan:
see here https://basecamp.com/3063126/projects/10728552/todos/219137373#comment_364793811

PR #1971, opened 2015-12-18
Updated carousel slide text with links

KarlaRenschler:
Added links and text for new carousel slide on Commodities mill points data.


apercas:
manually added

PR #1989, opened 2015-12-22
New UMD As It Happens layer

adammulligan:


aagm:
It gives me a 404 the first try, but at the second works very good

PR #2022, opened 2016-01-07
Cleaning. Ordering some code. Set tab when hresolution param is present

mbarrenechea:
- I have moved the function '_getParams' to the UrthecastLayer.js because it was only used by itself. 
- Set hresolution-tab opened when hresolution param is present.


mbarrenechea:
This isn't necessary anymore.

PR #2038, opened 2016-01-12
New user subscription workflow

adammulligan:
This updates the user subscription workflow so that user's can add names to their subscriptions. It matches the new designs, meaning that the subscription tab has been completely removed and user's now subscribe through the analysis tab and a modal.

You'll need the `user-api-improvements` API branch to test this locally.


apercas:
Superb.

PR #2043, opened 2016-01-13
added analysis for GLAD

aagm:


aagm:
Adam just left to check the dates retrieved from the timeline

adammulligan:
@aagm I think I've fixed this, are you able to try it out locally?

aagm:
this last fix solve almost every problem, just left to get the correct max date from the data and ready to go up

PR #2070, opened 2016-01-22
added mongabay stories

aagm:


mbarrenechea:
I'm going to merge it, but I think that it would be better if the GFW API handles it.

PR #2073, opened 2016-01-22
User flow improvements

adammulligan:
This fixes a bunch of little details in the user and subscriptions flows, such as fixing the "view on map" URLs, and adding validation to the profile form. The reason there's so much code here is because I returned the Subscription tab to the map :smiley_cat: 

It also fixes the main 404 page, to make it compatible with the new navbar (right now it's pointing to the staging assets, so we'll need to change that to master later).


app/assets/javascripts/map/presenters/tabs/SubscriptionPresenter.js
```diff
@@ -0,0 +1,500 @@
...
+      threshold: 30, // by default
+      iso: null,
+      overlay: null, // google.maps.Polygon (user drawn polygon)
+      multipolygon: null, // geojson (countries and regions multypolygon)
+      disableUpdating: false,
+      dont_analyze: false
+    }
+  });
+
+  var concessionsSql = {
```
apercas: I know it's just in the purpose of beauty but maybe we should use some URL templating here, what do you think? @adammulligan

adammulligan: Yeah, I think there's actually a lot in here to refactor. This file is actually quite old, but I'm just re-adding it after I mistakenly deleted it previously, so I haven't really changed anything. Perhaps it's worth us adding an issue here in the repo to refactor this soon? There's a ton of shared code between the map tab views that we could reduce down.

PR #2103, opened 2016-02-11
Disposable/metadata system

mbarrenechea:
Don't merge or deploy until we have the confirmation of @aagm. She has to update the layerspec with the new source_json column.


mbarrenechea:
After checking it seems to be correct. Let's deploy it. I'm talking with myself?

PR #2106, opened 2016-02-12
Implement the magical new geostore service

adammulligan:
We had an issue with uploaded shapefiles: the geojson was too damn ~~high~~ big for the URL params. As a solution, we now store geojson for analyses on the API, and store a reference to it in the URL params. Please upload some shpfiles, login for subscriptions, etc. and see if the state is correctly maintained! You can test this using the staging API, as I've deployed the code for that already.


mbarrenechea:
Great work Adam

PR #2152, opened 2016-03-11
Feature/asynch huginn

apercas:
This works and can be merged if necessary (e. g. as it was recently, so it won't take down the whole site) , but I will spend some more time cleaning it and making it prettier.

/cc @adammulligan @mbarrenechea


adammulligan:
Great work, thanks :star:

PR #2174, opened 2016-03-16
Disposable/little fixes

mbarrenechea:
Several little changes along the whole app.
- Map: Urthecast legend
- Map: Search box set a marker on the map whenever you make a search
- Global: typo fixes and other similar stuff


adammulligan:
Tested locally and all seems to be working :+1:

PR #2201, opened 2016-03-29
Fix/non compatible google fonts

apercas:
Updating local `.env` file to `GFW_ASSETS_URL=http://gfw-assets.s3.amazonaws.com/static/gfw-assets.latest.js` would be nice too :D


mbarrenechea:
I've just added the `font-feature-settings: "liga" 0;`. It would be great if we can check it in a computer with the font issue.

PR #2212, opened 2016-03-31
Feature/order video on stories

apercas:


adammulligan:
@apercas could you also modify it so that it removes the YouTube link from the "YouTube Video" input box when you delete the video from the media? Possibly vice-versa as well -- when you remove the YouTube link from the input, it removes the video from the media sections -- what do you think?

apercas:
@adammulligan tots, makes sense

PR #2213, opened 2016-03-31
Feature/fao amendments

mbarrenechea:
- Some typo changes
- Y-axis in burned forest graph

### DIFF
--- a/app/assets/javascripts/countries/views/CountryShowView.js
+++ b/app/assets/javascripts/countries/views/CountryShowView.js
@@ -472,14 +472,13 @@ define([
       var $el = $('.country-burned_forests');
       var $graph = $('.burned_forests-graph');
       var $comingSoon = $el.find('.coming-soon');
-      var json = _.filter(_.map($graph.data('json'), function(el) {
-        if (el.year == 2003 || el.year == 2004 || el.year == 2011 || el.year == 2012) {
-          return null;
+      var json = _.compact(_.map($graph.data('json'), function(el) {
+        if (el.year > 2004 && el.year < 2011){
+          return el;
         }
-        return el;
-      }), function(el) {
-        return (el != null)
-      });
+        return null;
+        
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/countries/views/CountryShowView.js
```diff
@@ -472,7 +472,14 @@ define([
       var $el = $('.country-burned_forests');
       var $graph = $('.burned_forests-graph');
       var $comingSoon = $el.find('.coming-soon');
-      var json = $graph.data('json');
+      var json = _.filter(_.map($graph.data('json'), function(el) {
+        if (el.year == 2003 || el.year == 2004 || el.year == 2011 || el.year == 2012) {
```
adammulligan: Are these specific years that don't have data, or are we trying to show only a certain range? If it's just a range of years where we only have valid data, then maybe it would be clear here to do something like `if (el.year > 2004 && el.year < 2011) { return el; } else { return null; }`

Also, it's not a necessary change, but you can use `_.compact` rather than `_.filter(..el != null..)`, which might be clearer, though possibly just personal preference.

mbarrenechea: I'm ok with the range part.

However, `_.compact` also removes 0 values. That's why I use filter

adammulligan: Ahh gotcha, I saw that they were objects so I didn't worry about the potential for 0s or other false values! Good point.

mbarrenechea: Ok, you were right. In this case `_.compact` doesn't affect the result, as long as they are objects.  It's fixed now

adammulligan: Awesome, thanks! Graph looks good and works well :+1:

PR #2222, opened 2016-04-05
Feature/explore map gallery

mbarrenechea:
- New map gallery added to explore page. 
- Filter functionality enabled
- Some style amendments
- Responsive


app/assets/javascripts/static/GalleryView.js
```diff
@@ -0,0 +1,144 @@
...
+
+    getCount: function(filter) {
+      var filteredCollection = (filter == 'all') ? this.toJSON() : this.filter(filter);
+      return filteredCollection.length;
+    },
+
+    filter: function(filter) {
+      return _.compact(_.map(this.toJSON(), function(v){
+        return (v.filter == filter) ? v : null
+      }))
```
adammulligan: This code is fine, though you can be a bit more idiomatic with regards to using Backbone's built in functions by replacing your custom filter function with something like: `this.where({filter: filter})`. This returns models, so if you want JS objects, you could use something like `_.invoke(this.where({filter: filter}), 'toJSON')`.

The more work we make Backbone do, and the less code we write, the better! :joy:

app/assets/javascripts/static/GalleryView.js
```diff
@@ -21,19 +21,20 @@ define([
...
-    filter: function(filter) {
-      return _.compact(_.map(this.toJSON(), function(v){
-        return (v.filter == filter) ? v : null
-      }))
+    customFilter: function(_filter) {
+      if (_filter != 'all') {
+        return _.invoke(this.where({filter: _filter}), 'toJSON');
+      }
+      return this.toJSON();
     },
```
mbarrenechea: I hope you will like this :sweat_smile:

adammulligan: Ahh so much better! :heart_eyes:

PR #2280, opened 2016-04-29
Feature/hres refactor

mbarrenechea:
We need to set a message/advice first time user enables the hr-layer. After that, I've set a cookie for 60 days to prevent this message appear again. Here you have the styles if you want to check that they are correct.
https://app.zeplin.io/project.html#pid=564c8b73362183187a126880&sid=57221fb360c90bd61e71f509

### DIFF
--- a/app/assets/javascripts/map/views/tabs/HighresolutionView.js
+++ b/app/assets/javascripts/map/views/tabs/HighresolutionView.js
@@ -204,7 +204,13 @@ define([
 
 
     printSelects: function() {
-      this.triggerChosen();
+      this.$selects.chosen({
+        width: '100%',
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/map/views/tabs/HighresolutionView.js
```diff
@@ -273,13 +281,6 @@ define([
     },
 
     triggerChosen: function() {
-      this.$selects.chosen({
```
adammulligan: Can you remove this empty function if it's not being used anymore?

adammulligan:
Ok, small comment on that empty function, but otherwise it works great locally ⭐

PR #2289, opened 2016-05-04
Disposable/hres

mbarrenechea:
- Some cleaning
- New styles
- Responsive datepickers
- Notification the first time you use it


apercas:
@mbarrenechea is `cookie` just a wrapper and handler of `document.cookie`?

mbarrenechea:
It's the js-cookie library that we were already using. So, answering your question, yes, I think that it's a wrapper of `documet.cookie`.
https://github.com/js-cookie/js-cookie
https://github.com/Vizzuality/gfw/blob/90918be507a9c39956f9375010b161a2c7b4c560/config/requirejs.yml#L39

apercas:
Cool

PR #2301, opened 2016-05-13
Country: module refactor, layers country dropdown.

mbarrenechea:
First of all, `bower install`. I have changed the version of chosen because the we were using a beta one. I've checked it along the app and I didn't find anything wrong. But feel free to make your own checks.

In this PR you will find:
- New country dropdown inside layersNavView ("LayersCountryView.js").
- Code related to countriesView refactored.
- Mobile version amendments. 
- Analysis & Countries view are totally separated, each view will control itself.
- Cleaning code (we should never find a mps inside a view. That's why we have presenters...)

Things that I think we should improve:
- There are still some hacks or patches (You will find comments in the code)
- Mobile list of letters don't disable the letters that don't have countries (i.e: there isn't any country with the "W" as first letter).
- The toggle of layers inside the "LayersNavView.js" or "LayersCountryView.js" are still being handled by the LayersNavView. I think that they should be separated even if we duplicate some code. As I've done with the "CountriesVIew.js"

@adammulligan keep calm and ask me whatever you want


app/assets/javascripts/map.js
```diff
@@ -74,26 +74,33 @@ require([
...
-      new GuideButtonView();
+
+      var map = new MapView();
+      new MapControlsView(this.map, this.countries);
+
+      this.map = map.map;
+      this.countries = new CountryCollection();
+
+      // Init views
+      this.countries.fetch().done(function(){
```
adammulligan: Although the amount of data being downloaded here is small, it does mean that the UI loading is dependent on CartoDB working and returning data in a timely manner. Maybe we should consider having the views below listen to the `sync` event on the collection. For example,

```
this.countries = new CountryCollection();
new TabsView(this.map, this.countries);
... etc ...
this.countries.fetch();
```

and then `TabsView` would have something like `this.listenTo(countries, 'sync', this.render)`. This way parts of the UI that don't rely on the countries can start rendering in the background.

Also, does every view listed here require the countries collection?

app/assets/javascripts/map/views/LayersCountryView.js
```diff
@@ -0,0 +1,158 @@
...
+      
+      this.cache();
+      this.chosen();
+      this.legibility();
+    },
+
+    cache: function() {
+      this.$select = this.$el.find('#layers-country-select');
+      this.$selectChosenId = '#layers_country_select_chosen';
+      // This belongs to the parent, is there a better way to do this?
```
adammulligan: Maybe we can move the `legibility` function to the parent view? If it is dependent on this view having rendered, you could potentially have this view trigger an event when it's rendered so the parent can resize, or inject a method from the parent.

app/assets/javascripts/map/views/LayersCountryView.js
```diff
@@ -0,0 +1,158 @@
...
+
+    render: function() {
+      this.$el.html(this.template({
+        countries: this.countries.toJSON(),
+        country: this.model.get('country'),
+        countryName: this.model.get('countryName') || 'Country',
+        countryLayers: this.model.get('countryLayers')
+      }));
+      
+      this.cache();
```
adammulligan: I'd like it if the method names could be a bit more descriptive, for example `cacheSelectors`, `fixLegibility`, `renderChosenSelector`, etc.

adammulligan:
I love the direction this is heading, nice work! I've left a few small comments throughout the code -- let me know if any need more clarification.

Do you want to fix the layer toggle handling you mentioned before I merge this?

I'm also very keen for us to start writing some tests on some of the GFW front-end. Maybe we can start on one of the more important methods you've introduced here, such as https://github.com/Vizzuality/gfw/pull/2301/files#diff-56ff8fd5ca8650ffbf659e4e0032fac7R277 and work together on writing a test for it over Skype? I think we'll get a lot of value out of writing a few small tests as we refactor, because then we can confidently change code in the future without worrying what side-effects we're causing.

app/assets/javascripts/map.js
mbarrenechea: No, not all the views need the country collection. Ok, I will try the sync of country collection to render the views.

app/assets/javascripts/map/views/LayersCountryView.js
mbarrenechea: You are right but it depends on the country update...(I've answered my question just right now). Yes, we should move it to the parent view and the layersNavPresenter should be subscribed to the Country/update event.

app/assets/javascripts/map/views/LayersCountryView.js
mbarrenechea: I don't agree... I want to standarize all the common view functions along the app ("cache", "render", "listeners"). Maybe you are right with the "fixLegibility". For example, you use "render" function and you don't name it "renderView".

For example "cache": we always cache selectors, but this function also cache variables that aren't selectors.

mbarrenechea:
About the layer toggle handling, I think that we can wait to fix it until we refactor the layersNavView and make it dynamic. (Raúl is working on it).

About the tests: of course I would like to do some of them with you via Skype.

app/assets/javascripts/map/views/LayersCountryView.js
adammulligan: Yeah, you're right about the `cache` name. I think my point stands regarding `legibility` and `chosen`, though -- these methods have side-effects so generally they should be verbs explaining the side-effects. `render` is a verb that explains what the function does, whereas `legibility` isn't

app/assets/javascripts/map.js
adammulligan: Ah ok, maybe in that case you can get away with depending on `countries.fetch()` but pull out some of the views that don't depend on the collection so that most of the UI loads while the countries are being fetched.

PR #2304, opened 2016-05-16
VIIRS for production

adammulligan:
@aagm the layer doesn't seem to be working now (or on staging), can you look in to it?

@mbarrenechea once alicia has got the layer working, can you test to make sure all is OK on the front-end and then deploy to production once Mikaela gives the go ahead? see here https://basecamp.com/3063126/projects/10726176/todos/231371569#comment_417848709


adammulligan:
@mbarrenechea @aagm ok, we're good to go to production once you've both verified the layer is working a-ok :-) https://basecamp.com/3063126/projects/10726176/todos/231371569#comment_418232410

PR #2310, opened 2016-05-18
this week’s GLAD update

manuela-stg:


adammulligan:
Thanks for the update! Unfortunately I'm getting an error from cartodb: `Pgraster Plugin: unable to lookup available table overviews due to unknown schema  encountered during parsing of layer 'layer0' in Layer`.

PR #2311, opened 2016-05-18
center timeline for all layers on big resolutions

apercas:
image


adammulligan:
Looks good for all layers, but I'm afraid it breaks the GLAD timeline:

screenshot 2016-05-18 16 30 34

apercas:
Weird, I checked and this is what I see:
image

PR #2316, opened 2016-05-19
biomass change colouring

aagm:
for staging and pro!

### DIFF
--- a/app/assets/javascripts/map/views/layers/PantropicalLayer.js
+++ b/app/assets/javascripts/map/views/layers/PantropicalLayer.js
@@ -32,7 +32,7 @@ define([
     filterCanvasImgdata: function(imgdata, w, h) {
       "use asm";
       // We'll force the use of a 32bit integer wit `value |0`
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/map/views/layers/PantropicalLayer.js
```diff
@@ -43,28 +43,30 @@ define([
...
+
           var intensity_scaled = myscale(intensity) |0,
-          bucket = (~~(countBuckets * intensity_scaled / 256) * 4);
+          bucket = (~~(countBuckets * intensity_scaled / 256) * 3);
 
-          imgdata[pixelPos] = c[bucket];
-          imgdata[pixelPos + 1] = c[bucket + 1];
-          imgdata[pixelPos + 2] = c[bucket + 2];
-          imgdata[pixelPos + 3] = c[bucket + 3];
+          imgdata[pixelPos] = 255-intensity;
```
mbarrenechea: I'm going to merge it, but it would be great if you could explain what is each pixelPos with some comments in another PR

PR #2327, opened 2016-05-25
Version removed. Nightly version is the correct one for develop

mbarrenechea:
@adammulligan @aagm @apercas @davidsingal @d4weed Please set this var in your .env file `GFW_ASSETS_URL=http://gfw-assets.s3.amazonaws.com/static/gfw-assets.nightly.js`. Since now we will always have the nightly version on local develop and staging. And the http://gfw-assets.s3.amazonaws.com/static/gfw-assets.latest.js` version should be the one set at production. 

Her we have a little documentation of how to use the gfw-assets: https://docs.google.com/a/vizzuality.com/document/d/1m_v-L7WVmDlBhd5eGhnpmHPTmaQ8c7gVvwUTaJweVv4/edit?usp=sharing

### DIFF
--- a/app/views/shared/_gfw_assets.html.erb
+++ b/app/views/shared/_gfw_assets.html.erb
@@ -18,8 +18,4 @@
 
   };
 </script>
-<script type="text/javascript">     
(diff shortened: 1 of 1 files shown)


adammulligan:
Thanks. Regarding the documentation, do you need to have a DOM element on the page already or does the script insert its own elements?

PR #2328, opened 2016-05-30
Terms of Service

mbarrenechea:
This Pr contains:
- New structure
- New behaviour (we don't navigate to the accept_terms anymore)
- Styles refactorized


mbarrenechea:
We are still working on this. We need to: "Warn users that they implicitly agree to GFW's terms whenever they download or upload data. i.e., "By downloading data you agree to the GFW Terms of Service [include hyperlink]." This message needs to be on the Open Data Portal (ideally surfaced at the time users click to download data), on the map when users download data, on the submit a story page, and when users upload data on the map for analysis and subscription." I'm going to talk with designers

mbarrenechea:
@adammulligan This is ready to merge. One thing:
- Eventually, we are not going to use the ConfirmModalView, but we can keep it for future implementations.

PR #2348, opened 2016-06-06
ArcGIS GLAD test layer

adammulligan:
I've replaced the existing GLAD layer with the new ArcGIS one, so that everyone can test it out on staging.

A few significant changes to note:

* I added the option of a tile loading spinner (like on Urthecast) to the `CanvasLayer`s, which only GLAD is using right now.
* I added a "layer options" URL param so that we can generically add custom toggle switches for layers


mbarrenechea:
Hey, everything seems to be working, but it's too slow...

PR #2349, opened 2016-06-06
Feature/country dropdown re amendments

mbarrenechea:
This PR contains:
- Analyze and subscribe from country dropdown
- New event to know if a layer is analyzable ('Analyze/enabled')
- Some standardisation of the mps events
- Legend fix


adammulligan:
I think we should maybe have some popover boxes for the analysis, etc. icons in the country dropdown as right now they're a bit confusing (I know they have the alt text, but it doesn't shop up straight away in most browsers). Perhaps something like the ones in the analysis box?

screenshot 2016-06-06 17 12 08

mbarrenechea:
Ok, I've added a tooltip but, if the text is bigger, we are going to have some issues. Let's see if this works for you

adammulligan:
No, that's perfect, thanks for changing it!

PR #2351, opened 2016-06-06
Sitemap implementation

andresgnlez:


app/assets/stylesheets/layouts/_sitemap.scss
```diff
@@ -0,0 +1,135 @@
...
+    margin-bottom: 15px;
+
+    > a {
+      font-size: 19px;
+      line-height: 1.7px;
+      font-weight: 300;
+      color: $cGreen;
+    }
+
+    &:hover {
```
mbarrenechea: It will better if we put the hover in the "a" tag, as long as the "li" is not going to take us anywhere. The same with the other ones

mbarrenechea:
Great!! I did the last little things, so don't worry about this anymore. Thank you

PR #2365, opened 2016-06-09
Fix: fit to geom fixed

mbarrenechea:
This PR contains:
- fix to geom if the param `fix_to_geom` exists
- console.log removed


adammulligan:
😍 😍 😍 

This seems to work in my quick local tests, but I think the real test will be trying it out on staging, etc. I'll deploy now!

PR #2367, opened 2016-06-09
SGF-amendments

andresgnlez:


app/views/small_grants_fund/index.html.erb
```diff
@@ -49,26 +49,26 @@
           <div class="frame js_frame">
             <ul class="slides js_slides">
               <li class="js_slide">
-                <div class="m-card">
-                  <a href="">
+                <div class="m-card -big-padding">
+                  <a class="full-click-link" href="">
```
mbarrenechea: I don't think that these are links...

app/views/small_grants_fund/index.html.erb
```diff
@@ -160,15 +160,15 @@
       <div class="m-card -fullWidth">
         <h2 class="card-title">Resources and support</h2>
 
-        <a href="" class="card-title-subitle _bold _uppercase">small grand funds webinars</a>
+        <a href="/getinvolved/apply-to-the-small-grants-fund" class="card-title-subtitle _bold _uppercase">small grand funds webinars</a>
```
mbarrenechea: This is incorrect. It should point to the new howto, As long as we are deploying it, I will change it, don't worry

app/views/small_grants_fund/index.html.erb
```diff
@@ -160,15 +160,15 @@
...
         <ul class="link-list">
-          <li><a href="">How to apply</a></li>
-          <li><a href="">Stories from the Field: 2015 projects</a></li>
+          <li><a href="/getinvolved/apply-to-the-small-grants-fund">How to apply</a></li>
```
mbarrenechea: You are already in the new page of small grants fund (the other one is going to be removed), this is not correct. Maybe you should set the Guidelines for Aplicants link. You have already usedit

PR #2462, opened 2016-07-12
Fix bug and move logic to presenter

j8seangel:
Fix bug not showing the modal after finish

Unify the logic into the presenter


app/assets/javascripts/map/helpers/guides.js
```diff
@@ -159,7 +146,7 @@ define([
 
       // Layers module
       {
-        selector: "#layersnav-forest-change",
```
j8seangel: We need to use the same editor config to avoid this kind of diffs (Mine is config to avoid trailing spaces)

adammulligan: I agree with this, I'm also not a fan of trailing white space. Maybe we should try something like http://editorconfig.org ?

j8seangel: We [have it](https://github.com/Vizzuality/gfw/blob/master/.editorconfig) !  So maybe it was an old file without the rules

PR #2468, opened 2016-07-19
Feature/subscription refactor

j8seangel:
Refactor the publish view:
- Put the logic together in the presenter.
- Use the new api
- Fix the user email update


app/assets/javascripts/map/presenters/tabs/AnalysisNewPresenter.js
```diff
@@ -349,10 +349,50 @@ define([
...
-          this.status.set('wdpaid', wdpaid);
+        'Analysis/shape': function(data) {
+          this.status.set({
+            useid: data.useid,
+            use: data.use,
+            wdpaid: data.wdpaid,
+          })
+        }
+      },
+      {
```
mbarrenechea: We will take a look at this function in the future. Too much logic in an mps subscription.

PR #2474, opened 2016-07-26
(Don't Merge Yet) New GLAD for production

adammulligan:
This is a disposable branch for merging the new GLAD layer to master, as we're not moving the new API to production in time for the August 2nd GLAD release. We might do a soft launch before the end of this week, but I'll need to find out from WRI first before we deploy this.

This introduces the following things that should be tested thoroughly before release:

* new GLAD layer with tiles automatically generated by WRI
* language selector for subscription emails
* GLAD timeline uses dates from ArcGIS
* Confirmed alerts toggle
* GLAD coverage layer should also contain Brazil

Before deployment, we also need to copy over the GLAD layerspec changes from nuclear_hazard to the production table.


mbarrenechea:
I'm going to close this one, as long as the new Glad layer will be deployed next monday

PR #2482, opened 2016-07-29
Fix/stories bugs

mbarrenechea:
This PR contains:
- Bug with location in Submit new story solved
- Bug with order in Submit new story solved
- Add credentials to Story Model


adammulligan:
:wave:

PR #2488, opened 2016-08-02
Osm basemap

mbarrenechea:
This Pr contains:
- New OSM basemap 
- Little improvements in the basemap list
- Little improvements in the wrapped layers. Use padding instead of margin to separate the options. i.e: Mexico land zoning


app/assets/javascripts/map/views/maptypes/openStreetMaptype.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * Dark Matter Maptype.
```
j8seangel: Is this description correct ?

PR #2492, opened 2016-08-09
Remove geostore/go from the place service

mbarrenechea:
Remove 'Geostore/go' from the place service.

It will only fire when a geostore is set on the URL. The one in charge to get the geojson is the AnalysisDrawPresenter, no one else needs it.


geriux:
Love PRs that remove lots of code =D 👍

PR #2529, opened 2016-08-31
Contribute data

mbarrenechea:
As long as I needed the same form styles I merged the branch `feature/subscriptions`

This PR adds 2 things:
- The work in the new subscription (don't worry, users can't access this form right now)
- The contribute data form with validations. (We still need to work on how to manage the uploads)


app/assets/javascripts/connect/views/MapMiniControlsView.js
```diff
@@ -0,0 +1,155 @@
...
+    initialize: function(map) {
+      this.map = map;
+      this.cache();
+      this.listeners();
+
+      this.renderGoogleAutocomplete();
+    },
+
+    listeners: function() {
+      this.status.on('change:is_drawing', this.changeIsDrawing.bind(this));
```
geriux: We should use 'listenTo' instead of 'on' so we make sure that when we call view.remove() those events will be removed as well =D

mbarrenechea: I didn't know that... Thank you!!! As long as I'm still working on this feature and it's not related to the PR, could we merge without changing this right now?

geriux: Yes @mbarrenechea let's do that!

PR #2537, opened 2016-09-02
Feat: Mails - Subscriptions confirmation

geriux:
- Added the key languages to the subscriptions
- Added mail templates for the subscriptions confirmation


app/assets/javascripts/map/templates/tabs/subscribe.handlebars
```diff
@@ -74,7 +74,11 @@
       <label for="subscriptionLanguage">Receive alert emails in...</label>
       <select data-placeholder="Select language" class="chosen-select default notranslate" id="subscriptionLanguage">
         <option value="en">English</option>
-        <option value="pt">Portuguese</option>
+        <option value="zh">中文</option>
+        <option value="fr">Français</option>
+        <option value="id">Bahasa Indonesia</option>
+        <option value="pt">Português (Brasil)</option>
```
mbarrenechea: These should be 'pr_BR' and 'es_MX', shouldn't they?

PR #2540, opened 2016-09-05
Contribute data: Upload files

mbarrenechea:
This PR allows users to upload files

### DIFF
--- a/app/assets/javascripts/static/ContributeDataView.js
+++ b/app/assets/javascripts/static/ContributeDataView.js
@@ -5,8 +5,9 @@ define([
   'underscore',
   'mps',
   'validate',
+  'jquery_fileupload',
 
-], function($, Backbone, Handlebars, _, mps, validate) {
+], function($, Backbone, Handlebars, _, mps, validate, jquery_fileupload) {
 
   'use strict';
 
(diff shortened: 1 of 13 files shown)


mbarrenechea:
image

app/assets/javascripts/static/ContributeDataView.js
```diff
@@ -81,6 +86,63 @@ define([
...
+
+    setUploadField: function() {
+      var that = this;
+
+      this.$fieldFileUpload.fileupload({
+        url: '/data/upload',
+        dataType: 'json',
+        autoUpload: true,
+        // acceptFileTypes: /(\.|\/)(gif|jpe?g|png)$/i,
+        maxFileSize: 10000000, // 10 MB
```
geriux: We should start using constants for things like the max file size, so it's easier to see if its at the top of the view.

app/assets/javascripts/stories/views/StoriesNewView.js
```diff
@@ -256,7 +256,7 @@ define([
             var confirmation = confirm('Are you sure?')
 
             if (confirmation == true) {
```
geriux: Let's try to use === 😁

geriux:
I think we should redirect the user to the home page or another section when the form is submitted, it gave me the impression that it didn't work, other than that it works great! 😁

PR #2581, opened 2016-09-20
Record country name on Google Analytics when there's an analysis on the map page.

simaob:


simaob:
Yeah you're right. I was too hopeful. =) Is there a best practice on this?
Do I just store the result of the findWhere on a var and then check for it's presence?

simaob:
Checked with Gerardo, and have made the change.

PR #2589, opened 2016-09-23
added uganda new PA layer

aagm:


aagm:
the metadata information is not working, seems that they havent addded it yet

mbarrenechea:
Ok, could you please tell them that they need to add it?

PR #2627, opened 2016-10-19
Add contact form to about page

j8seangel:
Use a new form in the about page instead the feedback form of the gfw-assets.

TODO: move the endpoint url to the new one when it is ready.

image

### DIFF
--- a/app/assets/javascripts/static/ContactUsView.js
+++ b/app/assets/javascripts/static/ContactUsView.js
@@ -208,7 +208,7 @@ define([
 
     validateInput(name, value) {
       let errors = validate.single(value, constraints[name]);
-      if (!!errors) {
+      if (!!errors && this.errors) {
         this.errors[name] = errors[0];
       } else {
         this.errors && this.errors[name] && delete this.errors[name];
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/static/ContactUsView.js
```diff
@@ -0,0 +1,244 @@
+define([
+  'jquery',
+  'backbone'
+], function($, Backbone) {
+
+  'use strict';
+
+  const topics = {
+    'report-a-bug-or-error-on-gfw': {
```
geriux: I tried to change the topic selector several times but the placeholder doesn't change, is that the expected behaviour?

app/assets/javascripts/static/ContactUsView.js
```diff
@@ -0,0 +1,244 @@
...
+      // Send request
+      this.$spinner.addClass('-start');
+      var xhr = new XMLHttpRequest();
+
+      xhr.open('POST', window.gfw.config.GFW_API_HOST + '/emails');
+      xhr.setRequestHeader('Content-Type', 'application/json; charset=utf-8');
+
+      xhr.onload = function() {
+        if (xhr.status === 200 || xhr.status === 201) {
+          this.changeStep('success');
```
geriux: Can we scroll to the div's top position? If you are close to the bottom of the page you might not see the success message.

PR #2685, opened 2016-12-15
fix/torque-player

pjosh:
* Fix a wrong behaviour with the player when you play with the dragger


geriux:
Hi! Thanks for the fix! There's still one more case we need to fix though:

Steps to reproduce:
- Click on play
- Drag the handle to the end (last date)
- Then if you click on play it doesn't play, if you click again (the stop icon) it will start playing with the wrong icon.

Can you please check it out? 

Thanks!

pjosh:
done @geriux! 🙏

geriux:
Yay! Awesome!!!! 💃 🎉

PR #2737, opened 2017-02-10
Fix the timeline end year value

pjosh:


geriux:
@pjosh Hey! let's check this tomorrow, had to remove it because it was breaking another layer. Let's discuss 😁

PR #2763, opened 2017-03-09
Feature/country pages v2 rails update

simaob:
This Pull Request upgrades the project to Ruby 2.4.0 and Rails 5.0.2.

You will need to have Ruby 2.4.0 installed, with the latest bundler, and then you should be able to run `bundle install` to get the updated gems.

During the upgrade I found some issues installing the Nokogiri gem, please refer to: http://www.nokogiri.org/tutorials/installing_nokogiri.html for information if it fails for you as well.

The project seems to work well after the upgrade. Let me know if you find any issues.

Have fun!


j8seangel:
Awesome and thanks for the update 🆕 🛤 🔝

PR #2810, opened 2017-04-11
added first basic automated test with reading from YAML file

benlaken:
Started to add basic framework for automated tests using `py.test` and made a simple first example. Instructions are in a README file within the newly added `/tests` sub-folder.

Essentially, install a few tools with Brew and pip as documented in the README, and then execute `py.test -v` from the root`gfw` folder.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -11,11 +11,9 @@ case "$1" in
         exec foreman start
         ;;
     test)
-        echo "Running Test"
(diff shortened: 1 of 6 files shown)


benlaken:
Looked into shifting to headless chrome with pytest, but it currently seems like GFW doesn't support it.

Test used: ` ChromeDriver 2.34.522932` and Selenium + Python 3.6

main-page

PR #2844, opened 2017-05-12
Feature/change slug

hectoruch:
Change:
- Slugs on glad and terra-i layers endpoint.
- Show region´s selector on glad and terra-i layers.


app/assets/javascripts/map/presenters/tabs/AnalysisNewPresenter.js
```diff
@@ -134,8 +134,8 @@ define([
         slug: 'imazon-alerts',
         subscription: true
       },{
-        name: 'terrailoss',
-        slug: 'terrai-alerts',
+        name: 'terrailoss', // change
```
geriux: Can we please remove these //change comments?

PR #2846, opened 2017-05-12
Add alert message on Glad layer

hectoruch:


hectoruch:
Add:
- New underline class on text styles.

Include:
- Settings and Text styles on map page.

PR #2850, opened 2017-05-17
Add new legend information

hectoruch:
- Add **_place to watch_** information legend template.

image


hectoruch:
Add:
- Legend icon.
- Create new column on carto database.
- Arrow icon.

image

PR #2855, opened 2017-05-18
Improve and add new styles on countries page detail

hectoruch:
- Improve header menu underline.
image

- Add new arrow icon on selects.
image

- Improve **"download data styles"** on header.

- Add new icons (file with all symbols)
image

- Improve styles on selects and graphs containers.


hectoruch:
Add:
- Colors on select icons.
- Improve ranking´s circles styles.

PR #2856, opened 2017-05-19
Feature/map countries page detail

hectoruch:
Add:
- Sticky map functions.
- New styles.

stickymap


hectoruch:
- Functions to center map on the country.

image

PR #2876, opened 2017-05-25
Include new features to tree cover loss alerts widget.

hectoruch:
- Active check data source filter.

checkdaasource


hectoruch:
Add:
- Functions to change source and origin filters.

showalerts

PR #2884, opened 2017-06-01
Feature/update widgets select region

hectoruch:
Add:
- Subscriptions function o all widgets.
- Change map when the region is selected.

countrychange


hectoruch:
Change:
- Tree cover loss ranking widget to update with a new region.
- Tree cover loss widget to update with a new region.
- Tree cover widget to update with a new region.

demo_2

hectoruch:
When region select change:
- Update map.
- Update all widgets.
- Change URL.

When user refresh the page with the new URL **_(/countries/IND/15)_**:
- Country selector change. (update handlebars).
- Map updates automatically.
- All widgets updates automatically.

hectoruch:
Fix all errors on URL text when user change the region or we choose all country data.

PR #2898, opened 2017-06-06
Change tooltip styles on map

hectoruch:
After:
image

Now:
image


hectoruch:
Change:
- Width styles.
- Image description styles.
- Image styles (on handlebars also change, now is a background.)
- Delete "Description" title.
- Create new Handlebars Function to limit strings on the description content and include three dots at the end.

image

PR #2899, opened 2017-06-06
Improve countries page.

hectoruch:
- Change legend styles with scroll functions.

changelegend


hectoruch:
legendchange2

hectoruch:
Fix (big errors):
- Now box legend does not disappear when region change.
- The loaders now disappear when region change.
- Delete console error when region change.

PR #2900, opened 2017-06-07
Improve and fix countries page map and other features.

hectoruch:
- Select first option change to "Whole Country" when the user changes the region. And when choose "Whole country" then change to "Select a jurisdiction".
- Fix errors when selecting another region, now the different layers appear.
- Improve map functions.
- Now the map update with animations and not re-create another new map for different region or country.

countrymapimprove


hectoruch:
Create a function to compare total cover loss value with the area of all the countries and get the closest.

image

image

PR #2912, opened 2017-06-13
Improve areas countries functions and include on widgets

hectoruch:
changearea


hectoruch:
Now on **Annual tree cover loss widget** on _View tree cover loss within section_ now appears comparison with random countries (every time we reload the page appears different countries.

aprox

PR #2920, opened 2017-06-13
applied earlier PR to remove 2015 highlight areas in UMD data

benlaken:
This merges changes from PR #2883 to remove 2015 highlighted areas from UMD. [Basecamp issue](https://basecamp.com/3063126/projects/10726176/todos/312158869).

### Before
BEFORE

### After 
BEFORE


app/assets/javascripts/map/cartocss/style.cartocss
```diff
@@ -61,6 +61,7 @@
   [layer='glad_coverage'],
   [layer='geographic_coverage_pantropical'],
   [layer='umd_coverage_2015'] {
```
simaob: we need to be careful with the code reviews guys. This broke Production. Fixing. (see the extra `{`)

PR #2938, opened 2017-06-21
Include reforestation rate ranking

hectoruch:
Add
- Ranking reforestation

Before:
image

After:
image


hectoruch:
Add:
- Current country on the ranking.
- Functions to know if the country is the first or the last one.
- Dynamic tree functions to draw it. (measure unit value)

image
### _(here the measure is 1000 and the first is the current one MONGOLIA)_

image
### _(here the measure is 100 and the Brazil (the current one) appear on the middle )_

hectoruch:
- https://www.pivotaltracker.com/story/show/147094069

PR #2941, opened 2017-06-22
Feature/line graph circle move

hectoruch:
Add:
- Now all the dots move at same time.
- Alerts values update at same time on all the cards.

ultimatelinegraph


hectoruch:
- Now when the region change, all works fine.

hectoruch:
- When the mouse is not on the graph, all the values restart to the last value.

linegraphmovemove

hectoruch:
- https://www.pivotaltracker.com/story/show/147518275

PR #2945, opened 2017-06-26
Fix zoom controllers

hectoruch:
- Add custom controllers on the map.
- Add a custom function to make zoom.
- Functions to see always the zoom control on top o bottom.

- https://www.pivotaltracker.com/story/show/147828185


simaob:
Looks good now!

PR #2948, opened 2017-06-26
Feature/add layer on map

hectoruch:
- Include new option on legend box.
- Order options on legend box.
- Include new layer.
- Change names.
- https://www.pivotaltracker.com/story/show/147827393

newlayer


hectoruch:
Merge https://github.com/Vizzuality/gfw/pull/2950

simaob:
Instead of using just the category color (forestChange for instance), could we use what we use on the map page, for the switches colors:

screen shot 2017-06-26 at 17 32 21

PR #2961, opened 2017-06-28
Feature/subscriptions workflow

pjosh:
subscriptions

If the user don't have `email` or `language` filled in his profile, we'll fill it with the subscription info.


simaob:
After this has been reviewed (by Hector) and merged to develop. Let's add it to the countries branch as well and deploy to staging.

PR #2976, opened 2017-07-17
Fix legends styles

hectoruch:
- Tooltip shadow.
- Change circle´s colors when the layer is hidden.
- Change cursor pointers on the eye icon and switch options.
- Change text color on switch options.
- Fix icon error on animation and colors.
- Change icon styles when the layer is hidden.

capture


hectoruch:
Fix:
- Circle styles on countries layers.
- Text legend on countries layers.

PR #2977, opened 2017-07-17
Fix land cover duplicate error

hectoruch:
image


simaob:
We'll fix this on the LayerSpec.

PR #2979, opened 2017-07-18
Fix countries selector on plantations option

hectoruch:
-Different functions for countries and the other layers.

fixcountryerrorlayer


hectoruch:
https://www.pivotaltracker.com/story/show/149061039

PR #2995, opened 2017-07-27
Update map menu.

hectoruch:


hectoruch:
- When the others menus or list are open, the header continues open always.

ezgif com-resize

PR #3001, opened 2017-07-28
Improve and fix styles on legend

hectoruch:
Fix&Improve:
-  Leave the metadata icons enabled even if the layer is made invisible.
- Make the legend elements gray as well when the layer is set to invisible.
- Add space between Category name and number of layers.

(https://www.pivotaltracker.com/story/show/149555321)


hectoruch:
- Remove parenthesis from the "tooltip" metadata

PR #3058, opened 2017-09-29
Feature/widget settings

pjosh:
## Overview

This PR include:
- Widget settings tooltip architecture.
- Architecture to update the widget.
- Widget layer when is loading (no styles).
- Tree cover settings (we still use the old endpoint so, only the canopy is affected)

## Demo

settings2

## Testing

- Visit the country page and test the tree cover settings.


hectoruch:
👍

PR #3062, opened 2017-10-03
Feature/improve widgets

hectoruch:
## Overview

This PR include:
- Improve sticky map. (real sticky interactions)
- Include real information on the header (forest data, forest-loss data).
- Include loaders on the page.
- Include loaders on widgets.
- Show more functions on widgets.
- Change arrow icons on select.
- Fix styles, and improve functions.

## Demo

- Sticky map -> Before:

beforemap

- Sticky map -> Now:

nowmap-iloveimg-compressed

- Loaders:

loaderpagewidget

- Real data on Header:

image

- More information:

showmore

- Arrow icon on selects:

iconselects

## Notes

(No notes)

## Testing

- Visit country pages and test it.

### DIFF
--- a/app/assets/stylesheets/components/country/_c-loading.scss
+++ b/app/assets/stylesheets/components/country/_c-loading.scss
@@ -1,8 +1,113 @@
 .c-loading {
-  height: 100vh;
+  height: calc(100vh - 59px);
   width: 100vw;
   display: flex;
   justify-content: center;
   align-items: center;
   align-content: center;
+
+  .loader {
+    font-size: 10px;
+    margin: 50px auto;
+    text-indent: -9999em;
+    width: rem(40px);
+    height: rem(40px);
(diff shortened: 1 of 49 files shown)


simaob:
On the gif that has the title `More information:`, I think that the ranking should update with the real positions of the states, which might mean we need more numbers and maybe more colours? Worth checking with Ben and Ari, but if we are seeing this:

1. Derp
2. Herp
3. Dord
...
10. Lerb


And we press down and there are other elements it means that those elements numbering should start at 11, because in reality the list is:

1. Derp
2. Herp
3. Dord
...
10. Lerb
(
11. Darp
12. Harp
...
x. Lorb

simaob:
Other than that comment it's looking good! =D

Ah... there's a loading spinner midway through the page, but maybe it's just waiting for something.

PR #3078, opened 2017-10-16
update content with carto

hectoruch:
## Overview

This PR include:
- https://www.pivotaltracker.com/story/show/151991607
- https://www.pivotaltracker.com/story/show/151991703

## Demo

newworld

## Notes

(No notes)

## Testing

(No testing)


j8seangel:
I think is not necessary to include a huge library like carto (652KB!)
image 
just to make an API request and fetch some data... so I'm going to remove it and do it by hand so we'll also drop another dependency (react-load-script)

Also, I've found a bug in the globe modal image so I'll fix it too.

j8seangel:
Same with THREE we are importing everything when we are only using a couple of functions! 
image

simaob:
When this PR is improved let's remember to update the feature/new-about-updated-webpacker to include the following two commits that are already in master:

- https://github.com/Vizzuality/gfw/commit/4c6e0b12851221b2bc6a445f1ddc4bdb09ba9d80
- https://github.com/Vizzuality/gfw/commit/5407336adb785f70da368be2b3c2bd4a02569bb3

PR #3093, opened 2017-10-31
Fixes issue with filtering data from past 2 days

simaob:
## Overview

There was an issue with the filtering of the forma active clearings for the past 48 hours, related to the timing of the alerts what we were hoping to do with the "Past 48 hours". We are now replacing that with "Past 2 days" and making sure that we capture the past two days and a bit more regardless of the time you visit the site.


benlaken:
I checked the returned data against Carto and it seems all is good to me :+1:
screen shot 2017-11-02 at 14 22 37
screen shot 2017-11-02 at 14 22 30

PR #3096, opened 2017-11-03
Fixed long-term problem with country rankings page

benlaken:
## Overview
As described in [this BC post](https://basecamp.com/3063126/projects/10726176/todos/327458495) there has been a problem with the country ranking page values (e.g. we have been saying that forest increased by >1,000% over the past 10 years), the values in the ranked lists were also incorrect: This PR addresses those issues.

(Below screen shots are for 50% tree canopy)
screen shot 2017-11-03 at 14 31 46
screen shot 2017-11-03 at 14 32 18

### DIFF
--- a/app/assets/javascripts/countries/views/CountryOverviewView.js
+++ b/app/assets/javascripts/countries/views/CountryOverviewView.js
@@ -228,10 +228,10 @@ define([
         }
 
         if (e) {
-          sql += 'OFFSET 10';
+          sql += ' OFFSET 10 ';
           $('.show-more-countries').fadeOut();
         } else {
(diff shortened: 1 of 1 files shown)


benlaken:
The total value and % value still needed some work, these have now been corrected.

* % value is now relative to the 2000 tree cover extent. I.e. the % is = (gain in ha / 2000 extent in ha)*100

* The absolute value matches the spreadsheet totals

Below screenshots are for 50% tree canopy thresholds:

screen shot 2017-11-03 at 17 03 58
screen shot 2017-11-03 at 17 03 53

PR #3099, opened 2017-11-06
Feature/fix glad range date

pjosh:
## Overview

Only the alerts from the last week (relative to the end date of the alerts) should be gold.


benlaken:
We need to use the date returned from this endpoint as the max date:
https://production-api.globalforestwatch.org/glad-alerts/latest
rather than the current date.

pjosh:
Hi @benlaken! I've used the max data date (the same as your endpoint), instead of the max date.
Regards!

PR #3104, opened 2017-11-13
Feature/2010 tree cover

01painadam:
Added the new 2010 Tree Cover Extent Layer to Land Cover menu.

- Selecting Tree Cover defaults to Tree cover 2000
- Tree cover 2010 Legend now has a drop-down selector, allowing you to choose 2000/2010 layer
- Both layers can be adjusted for Tree cover %
- Updated the subtitle for Tree cover in Land Cover menu

screen shot 2017-11-13 at 12 31 59
screen shot 2017-11-13 at 12 32 32
screen shot 2017-11-13 at 12 32 45
screen shot 2017-11-13 at 12 33 15


app/assets/javascripts/map/services/PlaceService.js
```diff
@@ -55,7 +55,7 @@ define([
   'use strict';
 
   var urlDefaultsParams = {
-    baselayers: 'loss,forestgain,forest2000',
+    baselayers: 'loss,forestgain,forest2000,forest2010',
```
mbarrenechea: Are you sure about this? I mean, it seems that you are setting by default both forest cover layers. Could you check it? Maybe I'm wrong...

01painadam:
Thanks @pjosh, yeah it is a little big (and not clickable everywhere) but it is in line with the current website in poduction if you look at the [Land Cover / Tree plantations legend](http://www.globalforestwatch.org/map/3/15.00/27.00/ALL/grayscale/none/670?tab=analysis-tab&dont_analyze=true).

I'll see if I can fix both of them!

pjosh:
Oh! sorry @01painadam, I've thought the selector was a new component. But thanks for the boy scouting ;)

PR #3132, opened 2017-12-05
Feature/location title

pjosh:
## Overview

Location selector only for Tree cover widget


app/javascript/pages/country/widgets/widget-tree-cover/widget-tree-cover-reducers.js
```diff
@@ -4,7 +4,7 @@ export const initialState = {
   totalIntactForest: 0,
   totalNonForest: 0,
   title: '',
-  locations: [],
+  indicators: ['gadm28', 'wdpa', 'ifl_2013'],
```
edbrett: I think maybe this should be in default props, as it wont change, doesnt have an action so doesnt belong in the store.

PR #3136, opened 2017-12-07
Fix/tree gain widget update

edbrett:
## Overview

Fixes for the Tree Cover Gain widget:
- Spread the data correctly from container to fetch and from state to props
- Correct loading state for fetch from countryData and component data
- Add check for fetch to make sure not already fetching
- Reduce component update logic so use minimal lifecycle methods

### DIFF
--- a/app/javascript/pages/country/root/root-component.js
+++ b/app/javascript/pages/country/root/root-component.js
@@ -16,24 +16,24 @@ const WIDGETS = {
   treeCover: {
     gridWidth: 6
   },
+  treeLocated: {
+    gridWidth: 6
+  },
+  treeLoss: {
+    gridWidth: 12
+  },
   treeCoverGain: {
     gridWidth: 6
+  },
+  treeCoverLossAreas: {
+    gridWidth: 6
+  },
+  totalAreaPlantations: {
+    gridWidth: 6
+  },
+  plantationArea: {
+    gridWidth: 6
   }
-  // treeLocated: {
-  //   gridWidth: 6
-  // },
-  // treeLoss: {
-  //   gridWidth: 12
-  // },
-  // treeCoverLossAreas: {
-  //   gridWidth: 6
-  // },
-  // totalAreaPlantations: {
-  //   gridWidth: 6
-  // },
-  // plantationArea: {
-  //   gridWidth: 6
-  // }
 };
 
 class Root extends PureComponent {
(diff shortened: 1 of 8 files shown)


app/javascript/pages/country/root/root-component.js
```diff
@@ -16,24 +16,24 @@ const WIDGETS = {
...
   treeCoverGain: {
     gridWidth: 6
-  },
-  totalAreaPlantations: {
-    gridWidth: 6
-  },
-  plantationArea: {
-    gridWidth: 6
   }
+  // treeLocated: {
```
simaob: why comment and not remove?

app/javascript/pages/country/widgets/widget-tree-cover-gain/widget-tree-cover-gain-actions.js
```diff
@@ -11,29 +11,28 @@ const setTreeCoverGainSettingsIndicator = createAction(
...
-        dispatch(setTreeCoverGainIsLoading(false));
-      });
+  params => (dispatch, state) => {
+    if (!state().widgetTreeCoverGain.isLoading) {
+      dispatch(setTreeCoverGainIsLoading(true));
+      axios
+        .all([getGain({ ...params }), getExtent({ ...params, threshold: 0 })])
+        .then(
+          axios.spread((getTreeGainResponse, getExtentResponse) => {
+            const gain = getTreeGainResponse.data.data[0].value;
```
simaob: I see that the data comes in a nice an meanigful structure =D

app/javascript/pages/country/widgets/widget-tree-cover-gain/widget-tree-cover-gain.js
```diff
@@ -56,48 +85,36 @@ class WidgetTreeCoverGainContainer extends PureComponent {
       item => item.value === settings.indicator
     );
     const regionPhrase =
-      settings.indicator === 'gadm28_only'
+      settings.indicator === 'gadm28'
```
simaob: would it make sense to keep a list of constants with the indicators names? to avoid many of these?

app/javascript/pages/country/widgets/widget-tree-cover-gain/widget-tree-cover-gain-actions.js
edbrett: Indeed. Not sure the decision in the API on this one.

app/javascript/pages/country/widgets/widget-tree-cover-gain/widget-tree-cover-gain.js
edbrett: They are currently in a constants, but this is the default. I will create a default location one and add on the next PR. Great idea.

PR #3140, opened 2017-12-08
Fix: locations widget update buttons

edbrett:
## Overview

Locations widget needed its fetching functions updating:
- Simple fetching and state management
- Manage loading correctly for tooltip and comp
- Correct header styles for widgets based upon size
- Disable buttons that are not functional

Additional:
- Reuse <SubNavMenu> comp with theme instead of tabs
- New component <Button> which handles all buttons now (accepts children)
- Separation of cards (styles were merging) with generic article card now available

## Notes

Again it will look like there are a lot of changes in there. This is mostly due to the about pages structure and lack of reuse of components before which now we have.


edbrett:
Ok sure. Can you show me which ones?

PR #3143, opened 2017-12-11
Real fetches for extent widget

edbrett:
## Overview

Updates to the extent service for extent widget:
- Get area along with value
- Fetch plantations and map values to reducer
- Remove unit from options list
- Reusable pie chart with initial settings for widgets (to avoid making lots of tooltips).
- Reusable <WidgetSettings> component that now decides what settings to show based on the props. HOOOORAY! No more extra components inside our widgets. 

EXTRAS:
- Small fix for sticky map when page is already scrolled


pjosh:
@edbrett, I made some changes to set the Location dropdown label as we need it.
If you are happy with my changes, you can merge this PR :)

PR #3150, opened 2017-12-13
Header sentence update

edbrett:
## Overview

We like things to be done. Done done done. Now the header sentence is done. Here's how it was done:
- Fetch many things (gadm28, plantations, area) from loss and extent services and then update store when done.
- Refactor sentence logic to handle all cases of null values from both loss and extent (check out Niger when done)
- Updates to the map to handle geostore loading in sync with the header when done.
- New loss service query to actually get the loss fetch done.

## Demo

screen shot 2017-12-13 at 12 05 20

## Testing

Done manually using my deep knowledge of select inputs and what the number should looks like. For confidence sake, check out Niger and Brazil in comparison at different juristictions.


app/javascript/pages/country/header/header.js
```diff
@@ -77,49 +80,91 @@ const mapDispatchToProps = dispatch =>
...
+          data.totalLoss.year
+        }, it lost <b>${lossWithOutPlantations}ha</b> of forest ${
+          data.plantationsLoss.area ? 'excluding tree plantations' : ''
+        }, equivalent to <b>${emissionsWithoutPlantations}tonnes</b> of CO₂ of emissions.
+      `;
     }
-  }
+    const totalSentence = `${firstSentence} ${secondSentence}`;
+
+    return <p dangerouslySetInnerHTML={{ __html: totalSentence }} />;
```
pjosh: What do you think about set the `dangerouslySetInnerHTML` on the component and this method only returns `{ __html: totalSentence }`. With this approach you don't need import `React` and we leave the responsibility of painting to the component.

edbrett: Yes I think you are correct.

PR #3167, opened 2017-12-19
Feature/embed widget

pjosh:
## Overview
* You need to add this to your `.env`: `GFW_URL=http://localhost:5000` (already on heroku)
* Embed strategy with a new country url
* Share component refactor
* Remove old svg icons sprite

## Demo

captura de pantalla 2017-12-20 a las 0 49 00


app/javascript/pages/country/widget/components/widget-header/widget-header-component.js
```diff
@@ -46,7 +39,7 @@ class WidgetHeader extends PureComponent {
               )}
             <Button
               className="theme-button-small theme-button-light square"
-              onClick={() => openShare(shareAnchor)}
+              onClick={() => openShare()}
```
edbrett: you only need openShare here.

config/routes.rb
```diff
@@ -116,8 +116,9 @@
 
   # countries
   get '/countries' => 'countries#index'
-  get '/country/*all' => 'country#index', as: :country
   get '/countries/overview' => 'countries#overview'
+  get '/country/embed/:widget/:iso(/:region)(/:sub_region)' => 'country#embed'
```
edbrett: Why can't we use the react router for this? Then we only need one entry point?

app/javascript/components/share/share-actions.js
```diff
@@ -1,11 +1,32 @@
 import { createAction } from 'redux-actions';
+import { createThunkAction } from 'utils/redux';
 
-export const setShareModal = createAction('setShareModal');
-export const setShareUrl = createAction('setShareUrl');
-export const setShareType = createAction('setShareType');
+import { getShortenUrl } from 'services/bitly';
+
+const setShare = createThunkAction('setShare', params => dispatch => {
+  const shareData = { ...params };
```
edbrett: You can spread all values you need as `const { embedSettings, embedUrl, ..... } = params.data`

app/javascript/components/share/share-component.jsx
```diff
@@ -12,42 +12,41 @@ import facebookIcon from 'assets/icons/facebook.svg';
...
           <input
             ref={input => {
               this.textInput = input;
             }}
             type="text"
-            value={url}
+            value={inputValue}
             readOnly
-            onClick={this.handleFocus}
+            onClick={() => handleFocus()}
```
edbrett: () not needed.

app/javascript/pages/country/embed/embed-component.js
```diff
@@ -0,0 +1,35 @@
...
+import WIDGETS_CONFIG from 'pages/country/data/widgets-config.json';
+
+class Embed extends PureComponent {
+  render() {
+    const { widgetKey } = this.props;
+
+    return (
+      <div className="row">
+        <div
+          className={`large-${WIDGETS_CONFIG[widgetKey].gridWidth} small-12`}
```
edbrett: I am not sure about including the grid inside the embed. It doesnt make sense to me. The widgets might need to be responsive? If so we need responsive iframes. Therefore we need to address all the styles and loading.

app/javascript/pages/country/embed/embed.js
```diff
@@ -0,0 +1,19 @@
+import { createElement, PureComponent } from 'react';
+import { connect } from 'react-redux';
+
+import EmbedComponent from './embed-component';
+
+const mapStateToProps = ({ location, countryData }) => ({
+  widgetKey: location.payload.widget,
+  isGeostoreLoading: countryData.isGeostoreLoading
```
edbrett: This is the only key that has no relevance to the widgets for loading. It is the country and region data we need often for the titles.

config/routes.rb
pjosh: because we have the header & footer loaded on rails, and the embed landing doesn't need it. So we need to load a specific view in rails.

app/javascript/pages/country/embed/embed.js
pjosh: But some widgets are using locationNames, so we need to wait to `isGeostoreLoading = true`, right?

app/javascript/pages/country/embed/embed-component.js
pjosh: Yeah, I don't like it either, but without that the styles doesn't work properly, so I decided to continue with the grid and later try to find a solution without the grid.
About the responsive, the widgets are 100% responsives.

app/javascript/pages/country/embed/embed.js
edbrett: geostore is only bound to the map so we dont need to wait for it. Countries, Regions, and SubRegions are what contains the data.

config/routes.rb
edbrett: Ah of course. Well done.

app/javascript/pages/country/embed/embed-component.js
edbrett: I cant seem to get them to be responsive when embedding. They dont fill the width of the container they are embedded inside of.

PR #3182, opened 2018-01-04
Feature/settings styles

edbrett:
Updated styles for the widget settings dropdowns and tooltip:
- Fonts on labels
- Label can have info icon
- Uniform pill dropdown style

screen shot 2018-01-04 at 17 35 30

screen shot 2018-01-04 at 17 35 43


app/assets/stylesheets/layouts/_countries.scss
```diff
@@ -217,7 +217,7 @@
 .tab-countries {
   background-color: #f1f1f1;
   ul {
-    max-width: 83.33333%;
+    max-width: 1080px;
```
j8seangel: Is this the max layout size?

I guess it is still there because some legacy code becuase the .row should be the one that fits that size... anyway could we get, at least, it from settings?

app/assets/stylesheets/layouts/_cover-title.scss
```diff
@@ -1,21 +1,29 @@
...
   background-repeat: no-repeat;
   background-image: image-url('countries/countries.jpg');
 
+  @media print, screen and (min-width: 850px) {
+    height: 300px;
+  }
+
   h2 {
-    max-width: 83.33333%;
+    max-width: 1120px;
```
j8seangel: same here

PR #3186, opened 2018-01-09
updated mangrove watch layer

benlaken:
## Overview

Finishes off the new Mangrove layer started in PR #2736 , and discussed in [BC thread](https://basecamp.com/3063126/projects/10726176/todos/294291962).

Layerspec has been changed accordingly.

screen shot 2018-01-09 at 12 18 12
screen shot 2018-01-09 at 12 18 26
screen shot 2018-01-09 at 12 18 33


benlaken:
Looks like there is a bug around whitelisting new layers with Forest change layers which we need to address too.

PR #3195, opened 2018-01-11
Feature/ui update

edbrett:
## Overview

Addresses a series of issues with the UI and styles of the page:
- Buttons and Drop-downs now have the correct GFW styles.
- Pie chart now has the label in the tooltip
- Dynamic sentence is the correct color
- New map button and layout for mobile

Also some code improvements have been made to make widget creation smoother:
- Move loader to parent (DRY)
- Create error handler inside widgets and header so that we can show the user how great the API is sometimes


j8seangel:
I like a lot the error handling, we should take this more in account!
Thanks

PR #3198, opened 2018-01-15
use extentYear like variable

pjosh:
## Overview

Use the extent year like variable in the sentence


benlaken:
Extent year is now highlighted :clap:

screen shot 2018-01-15 at 11 10 46

PR #3206, opened 2018-01-17
Added config for relative TC widget

benlaken:
## Overview

Added start of Relative tree cover widget by updating config file. This is the remaining piece of the tree cover widget which was split up after suggestions from Juan Carlos.

The widget should show a ranked list of relative tree cover (%), in a half-width widget.

The dynamic sentence should say. 

```<1st Sub region> had the largest relative tree cover of <X>%, compared to a regional average of <Y>%, considering tree cover extent in <year> with a canopy cover of greater than <Z>%.```

### DIFF
--- a/app/javascript/pages/country/data/widgets-config.json
+++ b/app/javascript/pages/country/data/widgets-config.json
@@ -43,7 +43,7 @@
     "active": true
   },
   "relativeTreeCover": {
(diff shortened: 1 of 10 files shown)


pjosh:
captura de pantalla 2018-01-17 a las 17 01 58

PR #3207, opened 2018-01-17
Feature/intact forest layer widget

edbrett:
## Overview

> "Change your opinions, keep to your principles; change your leaves, keep intact your roots."

-Victor Hugo

Oh boy, if Victor could have imagined the influence his work would have on our widget driven society today, maybe he wouldn't have caught pneumonia and would still be with us. 

Much like Victor's work, this new widget not only inspires, but gives a truly specific and detailed understanding of Intact Forests and their extent on a country and regional level. And if _hugo_ to a country that contains plantations, you can now explore how intact forests can be used to understand forest degradation within a region. Pretty powerful ey?! Thanks Vicky boy.

A breakdown of the poetic process is as follows:
- Copy the treeCover widget
- Change the fetches to handle dynamic plantations
- Update settings to limit user selection
- Change the extent calcualtions

## Demo

screen shot 2018-01-17 at 16 34 33

## Notes

I have the deepest more respect and understanding for the works of Victor Hugo. Little of the actual words, as I don't speak french.

### DIFF
--- a/app/javascript/pages/country/data/widgets-config.json
+++ b/app/javascript/pages/country/data/widgets-config.json
@@ -22,7 +22,6 @@
     "config": {
       "gridWidth": 6,
       "indicators": ["ifl_2013", "ifl_2013__wdpa", "ifl_2013__mining"],
-      "showIndicators": ["ifl_2013", "ifl_2013__wdpa", "ifl_2013__mining"],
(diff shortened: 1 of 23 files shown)


edbrett:
Ok this has now beed fixed. Going to merge so we can consolidate the primary forests.

PR #3248, opened 2018-02-02
stop repeating colors, fix bug with prop types

edbrett:
## Overview

- Remove widget color dependancy as it is now global
- Fix prop types bug

### DIFF
--- a/app/javascript/pages/country/widget/widgets/widget-fao-cover/widget-fao-cover.js
+++ b/app/javascript/pages/country/widget/widgets/widget-fao-cover/widget-fao-cover.js
@@ -2,7 +2,6 @@ import { createElement, PureComponent } from 'react';
 import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
 import isEqual from 'lodash/isEqual';
-import COLORS from 'pages/country/data/colors.json';
 
 import actions from './widget-fao-cover-actions';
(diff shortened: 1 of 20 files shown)


app/javascript/pages/country/widget/widgets/widget-fao-cover/widget-fao-cover.js
```diff
@@ -15,7 +14,7 @@ const mapStateToProps = ({ widgetFAOCover }, ownProps) => {
   const selectorData = {
     data,
     locationNames,
-    colors: COLORS.extent
+    colors: ownProps.colors.extent
```
pjosh: In some cases you are sending `ownProps.colors` to the selector, but in other cases you are sending the specific category, like `ownProps.colors.extent`. Is that ok?

edbrett: Yes that is correct. The goal was to use the widget type to dynamically send the colours to all widgets. In some cases the widget type doesn't line up as it is fao which is needed for the whitelist. Maybe you try to think of a better a solution. Or maybe it is a redundant PR.

pjosh: oh ok! I understand now :)
Looks good to me! good job.

PR #3250, opened 2018-02-02
feature/fao-deforestation

pjosh:
## Overview

FAO deforestation widget

## Demo

captura de pantalla 2018-02-02 a las 22 34 24

### DIFF
--- a/app/javascript/pages/country/widget/widgets/widget-fao-deforestation/widget-fao-deforestation-selectors.js
+++ b/app/javascript/pages/country/widget/widgets/widget-fao-deforestation/widget-fao-deforestation-selectors.js
@@ -1,0 +1,63 @@
+import { createSelector } from 'reselect';
+import findIndex from 'lodash/findIndex';
+import { format } from 'd3-format';
+
+const getData = state => state.data || null;
+const getLocation = state => state.location || null;
+const getLocationNames = state => state.locationNames || null;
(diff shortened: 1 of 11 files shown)


benlaken:
As far as I can see this widget still has a bug and is missing a key thing:

* Needs a settings menu, to enable users to select a year range (as described in the notebook)
* Year range should be described as a 5 year block, e.g. the label in the database will be 1990, which means 1990-1995. 
* Values should be shown for one (five year) block at a time, based on user selection.
* Some years may be returned with with `Null` data, we should hide those from the menu

pjosh:
sorry @benlaken I forgot to push my changes 🤦‍♂️

app/javascript/pages/country/widget/widgets/widget-fao-deforestation/widget-fao-deforestation-selectors.js
```diff
@@ -0,0 +1,67 @@
...
+        '.3s'
+      )(deforest)}ha/yr</b>${
+        humdef
+          ? `, of which <b>${format('.3s')(
+            humdef
+          )}ha/yr</b> was due to human activity`
+          : ''
+      }.`;
+    }
+    return `No deforestation data in <b>${currentLocation}</b>.`;
```
pjosh: @benlaken please, check this sentence when we don't have data.

PR #3294, opened 2018-02-14
Fix/whitelist for water bodies

edbrett:
## Overview

We have the unique case within the country pages that we are using gadm28 and fao data in order to find the countries list. This contains admins that are potentially bodies of water (and correctly so). However in the case of the country pages we dont want the user to be able to select these areas as the data is pretty useless in widget form. This PR adds a new blacklist of admin2 water bodies by ISO and admin1.

This is the reference table: https://wri-01.carto.com/tables/water_bodies_gadm28/public/map
You can use this for testing values.

### DIFF
--- a/app/javascript/pages/country/providers/country-data-provider/country-data-provider-actions.js
+++ b/app/javascript/pages/country/providers/country-data-provider/country-data-provider-actions.js
@@ -8,8 +8,6 @@ import {
   getFAOCountriesProvider,
   getRegionsProvider,
   getSubRegionsProvider,
-  getCountryWhitelistProvider,
-  getRegionWhitelistProvider,
   getCountryLinksProvider
 } from 'services/country';
 import { getGeostoreProvider } from 'services/geostore';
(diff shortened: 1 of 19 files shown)


edbrett:
One solution is to render the widgets after the whitelist fetches have completed.

edbrett:
I am confused. At the moment the widgets dont render until the whitelists have finished loading, so there shouldnt be a problem with intact forests and plantations.

benlaken:
Any chance we can extend this fix to cover one more edge case - the Caspian sea? 
`/country/XCA`

edbrett:
I can take a look. Can you give more detail on the issue? You just want The Caspian Sea to disappear? I am not sure humans were meant to have that kind of power.

benlaken:
In terms of humans having that power - we are doing pretty well with the Aral Sea! 
The Caspian sea is in the DB because its in glads28 as an admin 0 area:
http://www.globalforestwatch.org/country/XCA
But it doesn't make sense to have it as a country: it shouldn't be a selectable option from the admin-0 list.

edbrett:
I have made a fix for the whitelist issues. Please review again.

PR #3302, opened 2018-02-16
Feature/ga tracking for pages and actions

edbrett:
## Overview

For the react part of the application we have no analytics or any way of defining what events to track. This PR attempts to solve that. The core concept is to keep analytics and custom event as far as possible from the components. This is very hard to do. I will try to explain what decisions I have made and how to use this feature:

1. Add two custom functions to handle two types of google analytics events
- Page view: this tracks and change in the router and sends it away. It is imported and coupled to the redux-first-router think action
- Custom event: this is a middleware with redux that intercepts all actions, compares there name to that in a manifest of desired custom events, and then sends the relevant data to Google.

2. In order to send a custom event you must have an action. In all our cases the custom action we need comes from a button click. Therefore it is easy to create a non invasive adaptation of our button component that triggers and action when needed. This action also allows us to pass custom data through the button so that the analytics event can be curated.
Button gets pressed -> action is triggered -> payload is the data prop on the button -> follow same workflow as other actions 

3. Declarative custom events, when the keys you want to send can be callbacks, filters, strings, you name it! This allows for complete flexibility in how you decide to intercept an event and then send info to GA. 

## Demo

Hard to show, but try enabling react-ga debugger and watching the console.

## Notes

This is a little experimental, but hopefully opens up a pathway for better separation between analytics and project code.


app/javascript/pages/country/router.js
```diff
@@ -2,27 +2,32 @@ import { connectRoutes } from 'redux-first-router';
...
+
+  // track page with GA
+  handlePageTrack(getState().location);
 };
 
 export const routes = {
   [EMBED]: {
     path: '/country/embed/:widget/:country/:region?/:subRegion?',
-    thunk: syncWidgetSettings
+    thunk: routeChangeThunk
```
sorodrigo: Check the docs of `redux-first-router` there's a config option called `onAfterChange` that sets a thunk for all routes. I think it's better in this case.

app/javascript/pages/country/data/analytics.js
```diff
@@ -0,0 +1,62 @@
+export const ANALYTICS_EVENTS = [
```
sorodrigo: fancy 💅  !

app/javascript/utils/analytics.js
```diff
@@ -0,0 +1,74 @@
...
+    });
+
+    // process event if available
+    if (event) {
+      const eventPayload = {
+        category:
+          event.category && typeof event.category === 'string'
+            ? event.category
+            : event.category(payload),
+        action:
```
sorodrigo: I'd refactor this into a single function.
```js
const evalProp = prop => typeof prop === 'string' ? prop : prop(payload)
const eventPayload = {
...['category', 'action', 'label'].reduce((val, prop) => ({ ...val, [prop]: evalProp(event[prop]) }), {}),
...(!!event.value && { value: evalProp(event.value) })
};
```

There you go 😛

simaob:
@pjosh and @j8seangel would you mind reviewing this today? We are publishing some communications about the new country pages so would be great if we could have the events on production ahead of that.

j8seangel:
I love that idea !!! 👏  we could reuse it in other projects easily but there are a couple of things that potentially are tricky:
- The actions name and the analytics name doesn't come from the same origin so we should be very careful with those string... if any of them change we are not tracking anymore as for example: [here](https://github.com/Vizzuality/gfw/pull/3302/files#diff-f96a5b67a03b51d38ceb91997bd04599R3) and [here](https://github.com/Vizzuality/gfw/pull/3302/files#diff-dfceb8f988439a736b05ff586fc4645dR16) so having a const with the string could help.
- The `data` attribute doesn't seem very explicative to what is it doing... renaming it to `trackingData` or whatever could help to understand better what is it doing.

sorodrigo:
Regarding @j8seangel comments... Maybe instead of checking the action type like [here](https://github.com/Vizzuality/gfw/pull/3302/files#diff-3a79d2a483b2b35a774e3cb8fa337556R29), you could check for `action.meta.analytics.name`?

edbrett:
@sorodrigo if we do that then we have to customise all action to include that data, defeating the point of the abstraction.

edbrett:
@j8seangel the reason for 'data' was incase there was another reason to send data with a button at a later date. I was thinking of making a high order component to wrap things and pass the props so that the button props would need to be messed with.

edbrett:
I am going to merge as we need it on production and it does the job nicely. Any further comments let me know. I am going to try and improve this soon anyway.

PR #3303, opened 2018-02-16
Chore/software updates

simaob:
## Overview

This Pull Request upgrades both Nokogiri as well as Rails. There was a security vulnerability with the Nokogiri gem, and Rails released a new minor version, so we might as well update it! =)

## Testing

- get the branch
- run `bundle install`
- start the server
- everything should be the same as before.
- be proud of me. please. be.


edbrett:
This breaks the country pages dashboard for me, even after re installing dependancies, are you getting the same thing?

pjosh:
@simaob works good for me! @edbrett what's the error in your case?

benlaken:
Also works fine for me :+1: - going to merge

PR #3304, opened 2018-02-17
Feature/country zoom buttons

pjosh:
## Overview

- Map button theme.
- Zoom buttons on the map.
- Actions to manage the zoom with redux.

## Demo

captura de pantalla 2018-02-17 a las 1 35 34

### DIFF
--- a/app/javascript/components/icon/icon-component.jsx
+++ b/app/javascript/components/icon/icon-component.jsx
@@ -1,8 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import './icon-styles.scss';
-
 const Icon = ({ icon, className }) => (
   <svg className={`c-icon ${className}`} viewBox={icon.viewBox || '0 0 32 32'}>
     <use xlinkHref={`#${icon.id || icon}`} />
--- a/app/javascript/components/map/map-actions.js
+++ b/app/javascript/components/map/map-actions.js
@@ -4,10 +4,8 @@ import { fetchLayerSpec } from 'services/layer-spec';
 
 const setLayerSpecLoading = createAction('setLayerSpecLoading');
 const setLayerSpec = createAction('setLayerSpec');
-const setMapOptions = createAction('setMapOptions');
+const setMapSettings = createAction('setMapSettings');
 const setMapZoom = createAction('setMapZoom');
-const mapZoomIn = createAction('mapZoomIn');
-const mapZoomOut = createAction('mapZoomOut');
 
 const getLayerSpec = createThunkAction(
   'getLayerSpec',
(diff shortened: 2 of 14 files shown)


benlaken:
Works well for me and looks great :+1:

app/javascript/components/icon/icon-component.jsx
```diff
@@ -1,8 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import './icon-styles.scss';
```
edbrett: Is it not good to have some defaults?

app/javascript/components/map/map-actions.js
```diff
@@ -4,7 +4,10 @@ import { fetchLayerSpec } from 'services/layer-spec';
 
 const setLayerSpecLoading = createAction('setLayerSpecLoading');
 const setLayerSpec = createAction('setLayerSpec');
-const setLayers = createAction('setLayers');
+const setMapOptions = createAction('setMapOptions');
+const setMapZoom = createAction('setMapZoom');
+const mapZoomIn = createAction('mapZoomIn');
```
edbrett: I think you achieve this with a single action. changeMapZoom = (change) => state.map.zoom + change; and then set the buttons to pass +1 or -1? That way we can handle cases where we want to zoom by more than 1 step in the future.

app/javascript/components/map/map-reducers.js
```diff
@@ -1,8 +1,8 @@
 export const initialState = {
   loading: false,
   error: false,
+  options: {},
```
edbrett: I think we are starting to get messy with the map state. Options and settings to me are the same thing. Let chat to decide on how we should structure this.

app/javascript/components/map/map.js
```diff
@@ -81,6 +97,14 @@ class MapContainer extends PureComponent {
       stroke: '#333',
       fillColor: 'transparent'
     });
+    setMapZoom(this.map.getZoom());
+  }
+
+  setEvents() {
```
edbrett: Is this for when we change location and fitToBounds?

app/javascript/components/map/map.js
```diff
@@ -37,16 +38,25 @@ const mapStateToProps = (
 
 class MapContainer extends PureComponent {
   componentDidMount() {
-    this.buildMap(this.props);
-    this.props.getLayerSpec();
+    const { mapOptions, getLayerSpec, setMapOptions } = this.props;
+    this.buildMap();
+    getLayerSpec();
+    setMapOptions(mapOptions);
```
edbrett: I am not sure this is needed. If we are getting from the reducer then on mount it will already have them?

app/javascript/components/icon/icon-component.jsx
pjosh: yeah it's good if we had it, but we always overwrite all the values. I prefer to remove the useless code.

app/javascript/components/map/map-actions.js
pjosh: nice feedback! makes sense

PR #3311, opened 2018-02-20
Fix/remove glad cal legend

edbrett:
HOTFIX: remove 'detection date' legend from GLADs calendar & remove duplicated code causing undefined bug.


benlaken:
Slight wrinkle in that the remaining text under the widget says `Jump to Today` - this won't work for these data, so it should either become `Jump to latest date`, or be removed.

edbrett:
Currently not possible to change to jump to latest due to the calendar library. I have removed.

PR #3319, opened 2018-02-22
Feature/rails seo

edbrett:
## Overview

The country pages (and the whole app) needed a little SEO love. So we gave it. We are giving people. We give. To give is to gave. 

For this feature we have updated the SEO capabilities of the country page to include:
- Dynamic country fetch (courtesy of @simaob) to find the english country name from the ISO in the model.
- Server side dynamic rails SEO for both page sharing and widget sharing. Now if you share the page you recieve a title of `CountryName | GFW (ish)` and if you share a widget the social card reads `Widget Names in CountryName | GFW`. Pretty fancy eh?!
- Updated rails structure to allow react based layouts for embeds and pages without repeated code.
- General tidy of rails layouts and views

## Demo

Try sharing stuff with twitter.
screen shot 2018-02-22 at 11 32 57

## Notes

We still need to get this working with dynamic widget images. To plan is to use `dom-to-image` to allow use to snap shot the image and store it in an amazon bucket, where we will store it for 30 days or so.

## Testing

Going to put on staging and get posting.


app/controllers/country_controller.rb
```diff
@@ -1,14 +1,30 @@
...
 
-  layout 'country'
+  layout 'application_react'
+  before_action :set_country, only: [:show]
 
-  def index
-    @desc = 'Data about forest change, tenure, forest related employment and land use'
+  def show
+    @title = @country["name"]
+    @widget = params[:widget]
```
simaob: if you are not using a variable on the view you don't need to make it a instance variable `@variable` you can use just `variable`. Or in this case you can just use `params[:widget]`

edbrett: Ok will remove! Thanks for the review.

edbrett: Fixed :)

PR #3326, opened 2018-02-28
Feature: improved tooltips for components

edbrett:
Improves the use of tooltips in buttons and dropdowns. Now both can accept a much simpler tooltip object:
``` { text: "my excellent message" }```

However if you would like more customisation then you can pass any option from react tippy props as well:
``` { text: "my message", delay: 3000 }```

This was to reduce the importing of the Tip component throughout the project when its application was always the same.

In addition a tooltip has been added for the Dropdown component to allow the header section of the dashboard to have individual tooltips without extra code.

FINALLY, I also added a state prop which prevent tooltips from being shown in the header when a dropdown is open. This makes the user a lot happier I think.

### DIFF
--- a/app/javascript/components/button/button-component.jsx
+++ b/app/javascript/components/button/button-component.jsx
@@ -1,8 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Link from 'redux-first-router-link';
+import { isTouch } from 'utils/browser';
 
 import { Tooltip } from 'react-tippy';
+import Tip from 'components/tip';
 
 import './button-styles.scss';
 import 'styles/themes/button/button-light.scss'; // eslint-disable-line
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: app/javascript/pages/country/header/header-component.js [hunk 1]
```diff
@@ -1,20 +1,24 @@
...
 import arrowDownIcon from 'assets/icons/arrow-down.svg';
 import shareIcon from 'assets/icons/share.svg';
 import downloadIcon from 'assets/icons/download.svg';
 import './header-styles.scss';
 
 class Header extends PureComponent {
+  constructor(props) {
+    super(props);
+    this.state = {
+      disableTooltips: false
```
pjosh: good move!

app/javascript/components/button/button-component.jsx
```diff
@@ -68,7 +71,18 @@ const Button = props => {
...
 
   if (tooltip) {
-    return <Tooltip {...tooltip}>{button}</Tooltip>;
+    return (
+      <Tooltip
+        theme="tip"
+        position="top"
+        arrow
+        disabled={isDeviceTouch}
+        html={<Tip text={tooltip.text} />}
```
pjosh: Yeah, probably we will never use other component instead of `<Tip>` and if this happens we can overwrite it with `{...tooltip}`. I love it! good job.

PR #3331, opened 2018-03-06
Feature/gee recent imagery

pjosh:


app/javascript/pages/map/recent-imagery/components/recent-imagery-settings-component.js
```diff
@@ -0,0 +1,35 @@
...
+
+import draggerIcon from 'assets/icons/dragger.svg';
+import RecentImageryDrag from './recent-imagery-settings-drag';
+import './recent-imagery-settings-styles.scss';
+
+class RecentImagerySettings extends PureComponent {
+  render() {
+    const { settings, isDragging, connectDragSource } = this.props;
+
+    if (isDragging) {
```
edbrett: You could put this in the below return.

app/javascript/pages/map/root/root-styles.scss
```diff
@@ -0,0 +1,9 @@
+@import '~styles/settings.scss';
+
+.l-map {
+
+}
+
+.gm-style div div:nth-child(4) div:nth-child(4) div:nth-child(2) > div:not(.gm-style-iw) {
```
edbrett: Interesting! What does it do?

app/javascript/pages/map/router.js
```diff
@@ -0,0 +1,25 @@
...
+import queryString from 'query-string';
+import { handlePageTrack } from 'utils/analytics';
+
+const history = createHistory();
+
+export const MAP = 'location/MAP';
+
+const routeChangeThunk = (dispatch, getState) => {
+  // track page with GA
+  handlePageTrack(getState().location);
```
edbrett: I dont think we need this as the rails route will handle this in this case.

app/javascript/utils/map.js
```diff
@@ -0,0 +1,32 @@
+export const getPolygonCenter = polygon => {
```
edbrett: In this case, it might be worth using turf (you can import similar to lodash) http://turfjs.org/docs/#center. If this works then great, but for pease of mind a supported library might be more robust.

app/assets/javascripts/map/services/SentinelService.js
```diff
@@ -1,18 +1,18 @@
-define([
-  'Class',
-  'uri',
-  'bluebird',
-  'map/services/DataService'
-], function(Class, UriTemplate, Promise, ds) {
-
+/* eslint-disable */
```
edbrett: Do we need this service for the sentinel tiles inside the backbone app? Is this due to the react implementation?

edbrett:
Some thoughts/bugs:
- Dropdown's from selectors aren't appearing directly below input
- Calendar selector input has flashing cursor when clicked on (but not input)
- Can we limit the calendar to not go past the current month, as it is not needed?
- Clicking the 'i' button in the map layer legend doesnt open anything
- Hovering on the tiles doesn't show the text properly above it (blurry)

PR #3339, opened 2018-03-15
update rails country model to use gadm28 tables for country names

edbrett:
Update the fetches for the rails country helper to use the latest gamd28 tables instead of the old API (which doesn't exist).

### DIFF
--- a/app/models/country.rb
+++ b/app/models/country.rb
@@ -1,11 +1,11 @@
 class Country
   class << self
     def base_path
-      ENV['GFW_API_HOST']
+      "#{ENV['CARTO_API_URL']}/sql?q="
     end
 
     def find_all
-      url = "#{base_path}/countries"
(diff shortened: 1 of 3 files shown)


app/models/country.rb
```diff
@@ -22,12 +22,11 @@ def find_by_iso_or_name(value)
...
     def find_by_iso(iso)
       return nil unless iso
-      url = "#{base_path}/countries/#{iso.downcase}?thresh=30"
+      url = "#{base_path}SELECT%20iso,%20name_0%20as%20name%20FROM%20gadm28_adm2%20WHERE%20iso%20=%20'#{iso}'%20LIMIT%201"
       # CACHE: &bust=true if you want to flush the cache
       response = Typhoeus.get(url, headers: {"Accept" => "application/json"})
-
       if response.success? and (response.body.length > 0)
-        JSON.parse(response.body)
+        p JSON.parse(response.body)["rows"][0]
```
simaob: Do you really just want to print the result?

PR #3348, opened 2018-03-22
Feature/new analysis endpoint

pjosh:
## Overview

This PR changes the analysis url for the `umd-loss-gain` dataset to use the need endpoint that @01painadam made here: https://github.com/gfw-api/gfw-umd-forest-api/pull/8
I've also added a new select to analyze admin2 regions.

## Demo

admin2

### DIFF
--- a/app/assets/javascripts/services/CountryService.js
+++ b/app/assets/javascripts/services/CountryService.js
@@ -32,7 +32,7 @@ define(['Class', 'uri', 'bluebird', 'map/services/DataService'], function(
     getCountryConfig:
       "/query/{countriesConfigDataset}?sql=SELECT iso, indepth FROM {countriesConfigTable} WHERE iso='{iso}'",
     getCountriesList:
-      '/query/{countriesDataset}?sql=SELECT name_engli as name, iso FROM {countriesTable}',
+      '/query/{countriesDataset}?sql=SELECT name_engli as name, iso FROM {countriesTable} ORDER BY name',
     showCountry:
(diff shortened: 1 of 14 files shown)


app/assets/javascripts/services/CountryService.js
```diff
@@ -42,142 +50,217 @@ define([
...
-          },
-          error: function(errors) {
-            reject(errors);
-          }
-        };
-
-        this.abortRequest(GET_REQUEST_COUNTRIES_LIST_ID);
-        this.currentRequest[GET_REQUEST_COUNTRIES_LIST_ID] = ds.request(requestConfig);
-      }.bind(this));
+      return new Promise(
```
edbrett: Can we make sure these are alphabetically sorted?

edbrett: Currently Norway is first.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "globalforestwatch",
-  "version": "3.0.1",
+  "version": "3.1.0",
```
edbrett: Not sure why this is here?

pjosh: I don't know either 🤷‍♂️

simaob: 1 and 0 swapped places because 1 was tired of being on the window seat?

edbrett:
Additional comments:
- When there is no analysis data (Check norway for example) we should give feedback
- We are still missing the 2010 data in the analysis. Maybe you could discuss with JC how to handle this?

PR #3361, opened 2018-04-13
Move all components to global folder

edbrett:
Simple PR to move components that should be global to the components folder in order to simplify pages and reusability:
- Widgets
- Widget components
- Charts


edbrett:
I think you will need to restart your server

PR #3362, opened 2018-04-16
fix/map-timeline-mobile

pjosh:
timeline


edbrett:
I have made some extra changes to handle some bugs with the other menus also

PR #3368, opened 2018-04-20
Feature/decide update or reload

pjosh:
## Overview

We have dynamic update in the canvas layers, but when we need to change the threshold, it affects the url, so in this case we need to reload the layer.
threshold

### DIFF
--- a/app/javascript/components/map/assets/layers/loss.js
+++ b/app/javascript/components/map/assets/layers/loss.js
@@ -11,8 +11,8 @@ const OPTIONS = {
 };
 
 class Loss extends Canvas {
-  constructor(map, index, options) {
-    super(map, index, options);
+  constructor(map, options) {
+    super(map, options);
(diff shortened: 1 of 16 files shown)


app/javascript/components/map/assets/layers/loss.js
```diff
@@ -52,6 +52,13 @@ class Loss extends Canvas {
       .replace('{z}', z)
       .replace('{threshold}', this.options.threshold);
   }
+
+  setOptions(options) {
```
edbrett: Can this go on the parent layer?

pjosh: We have this method on the parent layer.
This is only an overwrite, for an specific behaviour for this layer.

edbrett: But it is repeated 3 times in loss, cover, and glads. When it could go in the parent canvas.js and still work fine?

PR #3370, opened 2018-04-20
feature/zoom-alert

pjosh:
## Overview
We need to show a zoom alert when the user has activated the Recent imagery tool.
The cancel button returns you to the previous zoom level.
zoom-alert


pjosh:
Of course we should! Little by little 😉

PR #3374, opened 2018-04-23
Biodiversity Significance Layer 3.0

01painadam:
## Overview

Oh boy, here I go updating Biodiversity layers again!

The new layer is possibly the greatest achievement in biodiversity layer visualisation. It's got **new technology**, implementing the RW-API layer hosting.

It's already being described as *on*, for production.

It also updates the layer to paint pixels only where forest data exists, deciding the significance of biodiversity in a region in a way that can satisfy any data scientist or researcher, at least. And indeed, *at most*.

Looking ahead to May, we can hope **not** to see any more updates to biodiversity metrics several times, and in various combinations.

## Note

Please leave open and merge with staging. I will need to update the LayersNav if everyone is happy with this layer to go to prod.


01painadam:
Now ready for PROD!

PR #3393, opened 2018-05-04
Feature/refactor components

edbrett:
Firstly, I apologise for the size of the changes in this PR. It was impossible to break this down further. 

This address DRY issues in the country page widgets, improving maintainability, performance of widget rendering, reduced fetching, and several fixes for property propagation and selector performance.

Please review with caution.


pjosh:
I've found a problem loading the glad layer on the country pages. I don't see any errors, but the layer doesn't load.

PR #3400, opened 2018-05-09
Feature/fao global

01painadam:
## Overview

Updates FAO reforestation and deforestation ranked widgets to be compatible with the global dashboard.
 
## Notes

- Both widgets likely need pagination on global view
- Fixes typo on reforestation widget

### DIFF
--- a/app/javascript/components/widgets/widgets/forest-change/fao-reforest/actions.js
+++ b/app/javascript/components/widgets/widgets/forest-change/fao-reforest/actions.js
@@ -8,7 +8,12 @@ export const getData = ({ params, dispatch, setWidgetData, widget, state }) => {
         (data.length &&
           data.find(d => d.iso === state().location.payload.country)) ||
         null;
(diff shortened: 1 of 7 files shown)


app/javascript/components/widgets/widgets/forest-change/fao-reforest/actions.js
```diff
@@ -8,7 +8,12 @@ export const getData = ({ params, dispatch, setWidgetData, widget, state }) => {
         (data.length &&
           data.find(d => d.iso === state().location.payload.country)) ||
         null;
-      dispatch(setWidgetData({ data: hasCountryData ? data : {}, widget }));
+      dispatch(
+        setWidgetData({
+          data: hasCountryData || params.type === 'global' ? data : {},
```
edbrett: Why not remove the country dependancy from the check above instead? So that you only check for data.length

01painadam: Good spot.

PR #3402, opened 2018-05-09
Feature/global gain ranked

01painadam:
## Overview

Prepares the *Ranked Gain* widget in the face of impending globalisation.

The bonus mayo on this pr-sandwich is that it also fixes a hither-to-unseen bug in the relative gain calculations, which also helped slim down the code in the `action.js` file.

## Notes

As always, please test.

This widget will also likely need pagination on the global page in the future.

### DIFF
--- a/app/javascript/components/widgets/widgets/forest-change/tree-cover-gain/actions.js
+++ b/app/javascript/components/widgets/widgets/forest-change/tree-cover-gain/actions.js
@@ -1,22 +1,23 @@
 import axios from 'axios';
 
-import { getGainRanked, getExtent } from 'services/forest-data';
+import { getGainRanked } from 'services/forest-data';
 
 export const getData = ({ params, dispatch, setWidgetData, widget }) => {
   axios
-    .all([getGainRanked(params), getExtent(params)])
+    .all([getGainRanked(params)])
     .then(
-      axios.spread((gainResponse, extentResponse) => {
+      axios.spread(gainResponse => {
         const gainData = gainResponse.data.data;
-        const extentData = extentResponse.data.data;
         let mappedData = [];
-        if (gainData && gainData.length && extentData && extentData.length) {
+        if (gainData && gainData.length) {
           mappedData = gainData.map(item => {
             const gain = item.gain ? item.gain : 0;
+            const extent = item.value ? item.value : 0;
             return {
               id: item.region,
               gain,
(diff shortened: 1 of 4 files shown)


app/javascript/components/widgets/widgets/forest-change/tree-cover-gain/actions.js
```diff
@@ -1,22 +1,23 @@
...
-      axios.spread((gainResponse, extentResponse) => {
+      axios.spread(gainResponse => {
         const gainData = gainResponse.data.data;
-        const extentData = extentResponse.data.data;
         let mappedData = [];
-        if (gainData && gainData.length && extentData && extentData.length) {
+        if (gainData && gainData.length) {
           mappedData = gainData.map(item => {
             const gain = item.gain ? item.gain : 0;
+            const extent = item.value ? item.value : 0;
```
edbrett: you can use `item.value || 0`

app/javascript/components/widgets/widgets/forest-change/tree-cover-gain/selectors.js
```diff
@@ -80,18 +77,25 @@ export const parseData = createSelector(
...
+    const {
+      initial,
+      withIndicator,
+      globalWithIndicator,
+      globalInitial
+    } = sentences;
     const locationData =
-      currentLocation && data.find(l => l.id === currentLocation.value);
-    const gain = locationData && locationData.gain;
+      currentLabel && data.find(l => l.id === currentLabel.value);
```
edbrett: currentlLabel is a string not an object. This should fail. Use `currentLocation` if you need the value

app/javascript/components/widgets/widgets/forest-change/tree-cover-gain/selectors.js
```diff
@@ -76,16 +77,24 @@ export const parseData = createSelector(
...
     if (!data || !data.length) return null;
     const {
       initial,
       withIndicator,
       globalWithIndicator,
       globalInitial
     } = sentences;
-    const locationData = currentLabel && data.find(l => l.id === currentLabel);
+    const locationData =
+      currentLabel && data.find(l => l.id === currentLocation.value);
```
edbrett: you are using currentLabel for checking. This should be the same as the comparison at the end of the line.

PR #3418, opened 2018-05-17
Feature/request cache

pjosh:
## Overview

This awesome feature prevents potencial issues when our external data sources fall.
Basically, we store the data source endpoints in a RedisDB. After that, we always try to get the information from Redis (much faster and reduces charge on the API).
The cache expires in 24h by default, but you can set other duration. Besides that, you can set cache exceptions if you need to cache the request but sometimes not.

## Test

It's fully integrated in the country pages. If you don't have Redis on you machine, run this command: ```brew install redis```


README.md
```diff
@@ -95,6 +100,13 @@ We follow a [Gitflow Worklow](https://www.atlassian.com/git/tutorials/comparing-
 
 ![gitflow workflow](https://www.atlassian.com/dam/jcr:b5259cce-6245-49f2-b89b-9871f9ee3fa4/03%20(2).svg)
 
+#### Clear Redis cache
+If you need to clear the Redis cache after deploy, run these commands on your local terminal:
```
edbrett: Can we automate this to happen on build? So when we deploy the cache is flushed?

edbrett: Well that is very annoying. maybe @simaob can help?

PR #3419, opened 2018-05-18
fix/ismobile-not-defined

pjosh:
## Overview

This fixes a JS error that we have since the last views refactor. I realize that we need the `isMobile` var defined on the `shared/js_footer`, but we have more stuffs inside it, so @edbrett, as you did the refactor, do you think this is a good idea? or we'll have duplicated things?

## Demo

captura de pantalla 2018-05-18 a las 10 10 18


edbrett:
No I think this works in this case! Good spot! Thank you!

PR #3430, opened 2018-05-24
Bugfix/nan loss location widget

benlaken:
## Overview

This PR address [the BC thread here](https://basecamp.com/3063126/projects/10727890/todos/350629881).

Specifically it:
* Catches cases where a NaN value was getting through on one of the widgets (e.g. Location of tree cover loss https://www.globalforestwatch.org/country/DJI?category=forest-change )
* Adds some extra sentences to several of the widgets to deal with no tree cover and no tree cover loss cases.
screen shot 2018-05-24 at 11 58 45
screen shot 2018-05-24 at 11 55 23

### DIFF
--- a/app/javascript/components/widgets/widgets/forest-change/tree-loss/initial-state.js
+++ b/app/javascript/components/widgets/widgets/forest-change/tree-loss/initial-state.js
@@ -39,7 +39,7 @@ export const initialState = {
     },
     sentences: {
       initial:
-        'From blah thing {startYear} and {endYear}, {location} lost {loss} of tree cover, equivalent to a {percent} decrease since {extentYear} and {emissions} of CO\u2082 emissions.',
+        'From {startYear} and {endYear}, {location} lost {loss} of tree cover, equivalent to a {percent} decrease since {extentYear} and {emissions} of CO\u2082 emissions.',
       withIndicator:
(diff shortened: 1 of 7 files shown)


juancarlosalonso:
Should we remove the rankings in these "no tree cover loss" cases?

benlaken:
It seems fine to leave it for now, people can use it as a navigator to different locations at least.

benlaken:
I should have mentioned yesterday that we should show the values if they are 0, so it is valid to show the list. The problem/question of whether or not to show the locations should only be if there was no data coverage in those locations (which is not the case).

app/javascript/components/widgets/widgets/forest-change/tree-loss/initial-state.js
```diff
@@ -39,7 +39,7 @@ export const initialState = {
     },
     sentences: {
       initial:
-        'From {startYear} and {endYear}, {location} lost {loss} of tree cover, equivalent to a {percent} decrease since {extentYear} and {emissions} of CO\u2082 emissions.',
+        'From blah thing {startYear} and {endYear}, {location} lost {loss} of tree cover, equivalent to a {percent} decrease since {extentYear} and {emissions} of CO\u2082 emissions.',
```
pjosh: this is a mistake, right? 😂

benlaken: lol my bad, I had a pretty high fever when I coded that  🤒

PR #3432, opened 2018-05-28
Forest type correction

01painadam:
## Overview

Changing indicator --> forestType was messing with the Intact & Primary Tree Cover Pie Chart Widgets.

- updates to forestType
- removes plantations segment from pie
- updates default year to 2010 in the case of primary forest

# Testing

Used IDN to check areas are consistent with main tree cover pie chart.


app/javascript/components/widgets/widgets/land-cover/intact-tree-cover/selectors.js
```diff
@@ -14,13 +14,11 @@ const getSentences = state => state.config && state.config.sentences;
 
 // get lists selected
 export const parseData = createSelector(
-  [getData, getSettings, getWhitelist, getColors, getCurrentLocation],
-  (data, settings, whitelist, colors, currentLabel) => {
```
edbrett: Do we still need the whitelist now?

PR #3441, opened 2018-05-30
Plantations loss

01painadam:
## Overview

Updates widget to use forestType instead of indicator.

### DIFF
--- a/app/javascript/components/widgets/widgets/forest-change/glad-ranked/selectors.js
+++ b/app/javascript/components/widgets/widgets/forest-change/glad-ranked/selectors.js
@@ -20,7 +20,7 @@ const getCurrentLocation = state => state.currentLabel || null;
 const getColors = state => state.colors || null;
 const getSentences = state => state.config.sentences || null;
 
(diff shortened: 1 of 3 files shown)


app/javascript/components/widgets/widgets/forest-change/glad-ranked/selectors.js
```diff
@@ -66,8 +66,7 @@ export const parseData = createSelector(
       return {
         id: k,
         color: colors.main,
-        percentage:
-          countsAreaPerc >= 0.1 ? `${format('.2r')(countsAreaPerc)}%` : '<0.1%',
```
edbrett: Do we definitely want to remove this?

01painadam: yes, it is unused. Although i've found another bug elsewhere so keep this open.

01painadam:
Also updated the Dynamic Sentence for GLAD ranking widget. The maths is consistent between both units now!

PR #3442, opened 2018-05-31
Feature - Low extent values

01painadam:
## Overview

All widget dynamic sentences should now report loss/gain/extent areas that are less than 1ha as a 3sf value instead of using symbol notation:

> e.g. 0.135 ha *instead of* 135 mha

## Testing

Please test all widgets - its difficult to find numbers in some cases! Most importantly, check that changing units (ha/%) still works as it should.

### DIFF
--- a/app/javascript/components/widgets/widgets/forest-change/fao-deforest/selectors.js
+++ b/app/javascript/components/widgets/widgets/forest-change/fao-deforest/selectors.js
@@ -60,6 +60,7 @@ export const getSentence = createSelector(
     const topFao = data.fao.filter(d => d.year === settings.period);
     const { deforest, humdef } = topFao[0];
     const totalDeforest = sumBy(data.rank, 'deforest');
+    const rate = currentLabel === 'global' ? totalDeforest : deforest;
 
     let sentence = humdef ? humanDeforest : initial;
     if (currentLabel === 'global') {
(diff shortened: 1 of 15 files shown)


app/javascript/components/widgets/widgets/forest-change/fao-deforest/selectors.js
```diff
@@ -70,10 +71,13 @@ export const getSentence = createSelector(
       location: currentLabel,
       year: period && period.label,
       rate:
-        currentLabel === 'global'
-          ? `${format('.3s')(totalDeforest)}ha/yr`
-          : `${format('.3s')(deforest)}ha/yr`,
-      human: `${format('.3s')(humdef)}ha/yr`
+        rate < 1
+          ? `${format('.3r')(rate)}ha/yr`
```
edbrett: A simpler way would be to create a const which find the '.3r/s' and then insert that into the format. But it isnt important.

app/javascript/components/widgets/widgets/forest-change/tree-loss-located/selectors.js
```diff
@@ -107,6 +107,16 @@ export const getSentence = createSelector(
     }
     const topLoss = percentileLoss / totalLoss * 100 || 0;
     let sentence = !indicator ? initialPercent : withIndicatorPercent;
+    const value =
+      topRegion.percentage > 0 && settings.unit === '%'
+        ? `${format('.2r')(topRegion.percentage)}%`
```
edbrett: Again here. You could do `const formatType = topRegion.percentage > 0 ? '.2r' : '.3s'`

PR #3443, opened 2018-05-31
Removes uncorrect Other Tree Cover values

01painadam:
## Overview

There seems to be an error in the FAO dataset (possibly rounding on their part) where the sum of all the forest types can be greater than the measured extent, resulting in this:

screen shot 2018-05-31 at 11 39 26

To avoid this, we now check that the sum of forest types **does not** exceed the extent. If it does, ignore it and set Other Tree Cover to 0ha.

If sum of forest types is less than total extent, Other tree cover is the remainder.

## Note

Also bundled in a new default setting for Tree Cover Ranking that was missed.


01painadam:
> uncorrect

PR #3455, opened 2018-06-01
Global 100%

01painadam:
## Overview

Makes pie chart on global dashboard add to 100% by correctly removing plantations areas (not included in calc)

### DIFF
--- a/app/javascript/components/widgets/widgets/land-cover/tree-cover/selectors.js
+++ b/app/javascript/components/widgets/widgets/land-cover/tree-cover/selectors.js
@@ -16,18 +16,18 @@ export const parseData = createSelector(
   [getData, getSettings, getWhitelist, getColors, getCurrentLocation],
   (data, settings, whitelist, colors, currentLabel) => {
     if (isEmpty(data)) return null;
-    const { totalArea, cover } = data;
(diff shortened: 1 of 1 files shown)


app/javascript/components/widgets/widgets/land-cover/tree-cover/selectors.js
```diff
@@ -16,11 +16,12 @@ export const parseData = createSelector(
   [getData, getSettings, getWhitelist, getColors, getCurrentLocation],
   (data, settings, whitelist, colors, currentLabel) => {
     if (isEmpty(data)) return null;
-    const { totalArea, cover, plantations } = data;
+    const { totalArea, cover } = data;
     const { indicator } = settings;
     const hasPlantations =
       (currentLabel !== 'global' && isEmpty(whitelist)) ||
       whitelist.indexOf('plantations') > -1;
+    const plantations = hasPlantations ? data.plantations : 0;
```
edbrett: why not get plantations from data still? If you are getting totalArea etc might as well do it the same way. Just being picky.

01painadam: See update.

PR #3456, opened 2018-06-01
Feature/paginate global gain widget

edbrett:
Adds pagination to the `Tree cover gain in` widget for the global case.


benlaken:
Looks great 👍

PR #3460, opened 2018-06-01
Feature/global loss ranking

01painadam:
## Overview

Adds a paginated, global version of the Loss ranking widget to the Global Dashboard 'forest change' tab.


benlaken:
Looking good 👍

PR #3462, opened 2018-06-04
Statistical Fires Widget

01painadam:
# Overview

Statistical fires widget! 🔥 
Does the same as current GLAD widget... but for fire.

For now I've just duplicated the GLAD widget as a starting point. We can likely reduce the code significantly at a later date.

**Cannot be implemented until we get the go ahead from Charlie that he's updating daily.**

For now, it looks like this:

screen shot 2018-06-05 at 12 28 27

image

***

## Notes

- Users are able to choose fire-type: ['MODIS', 'VIIRS']
- Adm0, 1, 2 available
- polynames available (if we want to use them. GLAD currently doesnt)

## Known Issues

- IOT, XCA, SPM, NFK, PIS, WLF require blacklisting

- reloading the page (e.g. moving between countries) doesn't always trigger the chart to refresh correctly

- start dates for each service/location *are not the same!* We will need a whitelist for which countries/services begin when.

- currently no api service to get 'latest' date. Once data is updated daily we should be able to do this using `viirs-active-fires` (minus 1 week to get the last full weeks worth of data)

- ~~'dataset' selector not appearing in user settings~~

- ~~data doesn't have zeroes when there are no alerts. That week/year is just missing.~~

- ~~requires a palette~~

- ~~there is a bug here where the statistical bounds in the background are not translated correctly, need to investigate~~

## Requirements

Update `.env` with:

```
    FIRES_ISO_DATASET=ff289906-aa83-4a89-bba0-562edd8c16c6
    FIRES_ADM1_DATASET=5bc5cd49-706f-409c-b10d-77fdfecb010f
    FIRES_ADM2_DATASET=0f24299d-2aaa-4afc-945c-b614028c12d1
```


app/javascript/components/widgets/widgets/forest-change/fires-alerts/component.jsx
```diff
@@ -0,0 +1,67 @@
+import React, { PureComponent } from 'react';
```
edbrett: As this is identical, could we import this into from the glads widget?

edbrett: Or even better make a new shared component in the `/widgets/components` folder?

app/javascript/components/widgets/widgets/forest-change/fires-alerts/selectors.js
```diff
@@ -0,0 +1,296 @@
+import { createSelector } from 'reselect';
```
edbrett: How much has this file deviated from the GLADs? Could we map the fires data to the same structure in the action and then make this a helper? Or even just reuse it from GLADs?

app/javascript/services/alerts.js
```diff
@@ -49,6 +54,22 @@ export const fetchGLADLatest = () => {
   return request.get(url, 3600, 'gladRequest');
 };
 
+export const fetchFiresAlerts = ({ country, region, subRegion }) => {
```
edbrett: Excellent.

app/javascript/components/widgets/widgets/forest-change/fires-alerts/selectors.js
01painadam: Yes, I think so - I just wanted to start from scratch first to find if there are many deviations from GLAD.

app/javascript/components/widgets/widgets/forest-change/fires-alerts/component.jsx
01painadam: Yeah we definitely could.

PR #3473, opened 2018-06-11
Feature: abstract caching helper to request level to make code simpler

edbrett:
This is a demo PR to test simplifying the cache methods for the react - redis caching set up. It does the following:
- Remove react caching provider and all reference to cache
- Fetches cache on a per request level
- Returns an additional promise inside the cache method to enable this


app/javascript/utils/request.js
```diff
@@ -5,35 +5,26 @@ import { getKey, addKey } from 'services/cache';
...
-    if (cacheError || cacheKeys.indexOf(key) === -1) {
-      const axiosInstance = axios.create();
-      const haveException = checkException(exceptionId);
-      if (!cacheError && !haveException) {
-        axiosInstance.interceptors.response.use(response =>
-          addKey(key, response.data, expire, exceptionId)
-            .then(() => response)
-            .catch(() => response)
-        );
+    return axios.get(`/cache/${key}`).then(cache => {
```
pjosh: With these changes we always do 2 request per request (cache + API) when the key is not saved (it happens because we don't have unlimited space). That's the reason of getting all the cache keys before.
Of course, we can live with that, but it goes against performance directly.

edbrett: Ah that was not the goal! The idea was that for each fetch you attempt to get that cache data (using the request as the key). If there is cache then it simply returns and you have the data, if not it gets the data and caches it. I wanted to try it this way to test performance as it might be a couple more fetches but they are very very small fetches only in the case that there isnt cache. And then we dont have to worry about bugs in the `componentWillRecieveProps` functions of each component which scare me. Now caching is separate from the components. I might have made some mistake in this change so let me check again.

PR #3475, opened 2018-06-11
Feature/conditional extent year option

edbrett:
When viewing IFL_2013 data we must use 2010 extent data. This forces the change when the user selects ifl_2013 and also makes sure the selector is hidden in that state.

BONUS: remove TWN and XCA from countries service as it is urgent.


pjosh:
The swap to 2010 extent data when selecting `ifl_2013` works well! But I don't know what you mean with `also makes sure the selector is hidden in that state`

edbrett:
When you have 'ifl_2013' selected from the frestType menu, the extentYears selector becomes hidden as it is not allowed to be changed.

PR #3483, opened 2018-06-15
LCC widget: updates calculation

01painadam:
## Overview

Reworks the maths... but still gets the same result


edbrett:
Conflicts in this. Also can you remove the uneeded second data file ?

PR #3488, opened 2018-06-18
Feature/lcc widget

pjosh:
## Overview

captura de pantalla 2018-06-18 a las 16 46 49


app/javascript/components/widgets/widgets/land-cover/global-land-cover/initial-state.js
```diff
@@ -0,0 +1,25 @@
+export default {
+  title: {
+    withLocation: 'Land cover for {location}'
+  },
+  config: {
+    size: 'small',
+    categories: ['land-cover'],
+    admins: ['country', 'region'],
+    selectors: [],
```
edbrett: no need for empty state.

app/javascript/components/widgets/widgets/land-cover/global-land-cover/initial-state.js
```diff
@@ -0,0 +1,25 @@
...
+    years: [2015],
```
edbrett: If no selectors is this needed?

app/javascript/services/forest-data.js
```diff
@@ -39,16 +39,17 @@ const SQL_QUERIES = {
...
 };
 
 const getExtentYear = year =>
   (year === 2000 ? 'area_extent_2000' : 'area_extent');
 
 const getLocationQuery = (country, region, subRegion) =>
-  `${country ? `iso = '${country}' AND` : ''}${
-    region ? ` adm1 = ${region} AND` : ''
-  }${subRegion ? ` adm2 = ${subRegion} AND` : ''}`;
+  `${country ? `iso = '${country}'` : ''}${
```
edbrett: This is used by a lot of queries, and subsequently changing it has broken other widget fetches. See tree loss for Brazil widget. This needs to be reversed.

app/javascript/components/widgets/widgets/land-cover/global-land-cover/initial-state.js
pjosh: Yes it's needed because we're using it to build the dynamic sentence.
AJ wrote in the notebook that we'll need to use a years selector, but is WIP now on the API side.

edbrett: Ok good to know. Thanks!

app/javascript/services/forest-data.js
edbrett: After further review this is a special case compared to all other queries. In the global widget cases we sometimes don't want 'AND' after {location}. However in this case we need it separated from the sql. Tricky.

pjosh: Oh I didn't see any widget broken. Let me check again!
Anyway, we need to adjust this logic because print `AND` at the end of all cases doesn't have much logic because if the following conditions are not met, it breaks.

app/javascript/components/widgets/widgets/land-cover/global-land-cover/selectors.js
```diff
@@ -0,0 +1,72 @@
...
+        value: sumBy(
+          dataFiltered.filter(o => d.classes.indexOf(o.key) > -1),
+          'value'
+        )
+      };
+    });
+    const dataParsed = dataMerged.filter(el => el.value !== 0).map(el => ({
+      ...el,
+      percentage: 100 * el.value / total,
+      value: el.value * 300 * 300 / 1e4,
```
01painadam: Just went back through the notebook and realised that (300 * 300 / 1e4) is unnecessary since my pipeline already converts pixels to Ha.

This is my bad, sorry!

app/javascript/services/forest-data.js
edbrett: For the case of global widgets fetches it is necesary, as we dont want to send a location so we need to remove the AND either at the beginning or the end.

pjosh: Fixed!

PR #3490, opened 2018-06-19
GFW Map 3.0

pjosh:
# Welcome to the GFW Map 3.0
9tdm4zl CREDIT: https://hotpapercomics.com/.

This feature branch brings with it a complete redesign and rebuild of the GFW map. Using the new react-redux architecture, [layer manager](https://github.com/Vizzuality/layer-manager), resource watch's [wri-api-components](https://github.com/resource-watch/wri-api-components), and the Resource Watch API itself, we have rebuilt the map to be lighter, faster, and more maintainable for all. 

Core features and improvements: 
- New layers menu and categories
- Improved legend with dataset specific timelines inside
- Improved analysis workflow with new one click analysis feature
- New EXPLORE menu which also users to view presets for the map to help them access curated content
- New SEARCH menu which also quick location finding and deeper dataset discovery
- Separated map controls section grouping base maps, recent imagery, and map view controls for better understanding
- Multi dimensional base map and label controls allowing users to be more specific about which boundaries, base map and label combinations they want.
- Brand new mobile interface and workflow for all sections of the map

Many thanks to all involved, you know who you are.

### DIFF
--- a/app/javascript/components/map-v2/components/map-tour/component.jsx
+++ b/app/javascript/components/map-v2/components/map-tour/component.jsx
@@ -256,7 +256,7 @@ class MapTour extends PureComponent {
           styles={{
             options: {
               overlayColor: 'rgba(17, 55, 80, 0.4)',
-              zIndex: 1000
+              zIndex: 2000
             }
           }}
         />


config/routes.rb
```diff
@@ -94,10 +94,11 @@
...
   # map
-  get '/map' => 'map#index'
-  get '/map/*path' => 'map#index'
-  get '/embed/map' => 'map#embed'
-  get '/embed/map/*path' => 'map#embed'
+  get '/map' => 'map_legacy#index'
+  get '/map/*path' => 'map_legacy#index'
+  get '/embed/map' => 'map_legacy#embed'
+  get '/embed/map/*path' => 'map_legacy#embed'
+  get '/v2/map' => 'map#index'
```
edbrett: Can we call this /map/v2 ? It just doesnt seem right this way around 😁

benlaken: fine with me

PR #3505, opened 2018-06-29
Feature/datasets provider

edbrett:
Starting point for the data and analysis menu on the map. 
- Add datasets provider from RW API.
- Create wrapper component for data and analysis tab with tab navigation.
- Add wri legend component for showing all active layers
- Allow all sub route for new map


app/assets/javascripts/map/services/GeostoreService.js
```diff
@@ -1,100 +1,88 @@
...
         ds.request(requestConfig);
-
-      });
+      }));
     }
-
   });
 
   return new GeostoreService();
-
```
pjosh: These changes breaks IE

app/assets/javascripts/map/views/layers/LossLayer.js
```diff
@@ -3,97 +3,106 @@
...
+          threshold: this.threshold
+        });
+      }
+    });
 
-  return LossLayer;
-});
+    return LossLayer;
+  }
+);
```
pjosh: Same here

edbrett:
Good stop! I didnt notice those. Fixed. Igtm.

PR #3508, opened 2018-07-03
Feature/map routing

edbrett:
## Overview

This is a big one. Both in file changes and architecture, so please review with an open mind and with caution!

So the new map we are going to need to manage a large amount of state specific to many components on the page. Additionally this map needs to be able to exist in the map page, dashboards, and potentially other forms and pages. So how do we make sure we can manage the map state robustly, in isolation, and not have to faff around with lots of url and location functions every time we use the map. Simple! We use the same encoding method we have been using for the widgets, but with some updates:

- we create a `?map=state` key in the url and bind it to the url by creating a single `setMapSettings` action to dispatch to redux-first-router.
- we move the previous decode and encode url to helpers that get triggered by the `querySerializer` of the router
- we check inside the `querySerializer` encoder to make sure what we are dealing with is an object and then encode it and send to the url
- the `location` key on the store is now holding the state of each component under a suitable key which can easily be passed to the component with its mapStateToProps and selectors.

This approach removes the need to dispatch an additional action to sync the url with the store like before, improving performance, and reduces complexity in the router when wanting to include a new component in a page.

Disadvantage: unreadble urls.

NOTE: I have left in some commented functions incase this approach is not approved to that we can comprise on some features.

## Demo

Open up redux tools, and try the zoom buttons on the map.

Credit: pretty much @j8seangel's PR.


app/javascript/components/map/map.js
```diff
@@ -44,8 +50,8 @@ class MapContainer extends PureComponent {
   }
 
   componentWillReceiveProps(nextProps) {
```
j8seangel: This will be deprecated soon, why don't use componentDidUpdate instead?

j8seangel:
Loving the structure selector! Feel free to merge and thanks!

PR #3510, opened 2018-07-12
changed GLAD tiles url on staging

benlaken:
Potential change to GLADS tiles url coming. This branch is for testing, and adding changes to staging. See [BC thread here](https://basecamp.com/3063126/projects/10728552/todos/356508574#comment_634281118).


edbrett:
This has now been handled dynamically based on env, so I am going to close. Great stuff.

PR #3512, opened 2018-07-16
Feature/remove widget dependencies

pjosh:


app/javascript/components/widgets/components/widget/widget.js
```diff
@@ -53,7 +52,9 @@ const mapStateToProps = ({ widgets, location }, ownProps) => {
   const locationPath = `dashboards/${type || 'global'}/${locationUrl}`;
   const widgetQuery = `widget=${widget}`;
   const widgetState =
-    query && query[widget] ? `&${widget}=${query[widget]}` : '';
+    query && query[widget]
+      ? `&${widget}=${btoa(JSON.stringify(query[widget]))}`
```
edbrett: Ah ha! You should no longer need to do this. You can now find this state as a clean and tidy object on the `location.query.widgets` reducer in the store. You can use a selector to merge the props just like in the map!

pjosh:
@edbrett the FAO widget has failed because this branch was outdated with develop and had the ` 1 = 1` sql issue. Now it's fixed!
About the Loss widget, yeah it fails, but the issue was in other PR that we already merge 🤦‍♂️ . I've fixed it directly in `feature/map-v2` https://github.com/Vizzuality/gfw/pull/3490/commits/6092880a345cc2e1888606d2cd3065a8307cd072 and now is up to date here.

app/javascript/components/widgets/components/widget/widget.js
pjosh: I don't do it to update the current url. Actually it's only for generate the share url that we need to pass to the `EXPLORE ON GFW` button on the embed view.

PR #3513, opened 2018-07-17
feature/analysis-location-selects

pjosh:
## Overview

kapture 2018-07-17 at 16 41 29


app/javascript/pages/map/data-analysis-menu/component.jsx
```diff
@@ -67,6 +69,8 @@ class DataAnalysisMenu extends PureComponent {
               analysis.showResults && <PolygonAnalysis />}
           </div>
         )}
+        <CountryDataProvider location={analysis.location} />
```
edbrett: We are likely going to need country data outside of the analysis on the map in many cases. Think embed forms etc. We could leave it here, but maybe it is clearer at the page level?

PR #3523, opened 2018-07-20
[ANALYSIS]: Fixes loss & multiple analyses

01painadam:
## Overview

The hottest, dirtiest fix you'll see this year - _guaranteed!_

- Fixes loss for country/adm shapes, custom shapes, wdpa, tcl - you name it!
- Allows multiple layers to be analysed simultaneously in the above cases

**Note**: one issue remains...

When analysing GLAD, the api currently only returns 2000 extent, so switching between 2000 & 2010 is broken. Requires multiple fetches in this instance... but thats for another PR.

### DIFF
--- a/app/assets/javascripts/map/services/AnalysisNewService.js
+++ b/app/assets/javascripts/map/services/AnalysisNewService.js
@@ -190,14 +190,13 @@ define(
                           }
                         };
                       } else {
-                        console.log('RESPONSE', response);
                         data = {
                           data: {
                             attributes: {
                               areaHa: response.data.attributes.areaHa,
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/map/services/AnalysisNewService.js
```diff
@@ -86,27 +86,51 @@ define(
                             }
                           }
                         });
+                        console.log('UMD', umdResponse);
```
edbrett: Can you clean up these

01painadam: done!

PR #3534, opened 2018-07-31
Feature/datasets connected to menu

edbrett:
An unacceptably large PR to connect the mapMenu to the datasets from the API and also to the map. Key features:
- Add leaflet map
- Isolate the map component
- Update `MapMenu` component to match the desired API structure
- Performance updates to selectors and toggle switch


benlaken:
Works locally for me, and is looking really good!

PR #3537, opened 2018-08-02
Adds simplify to geostore requests

01painadam:
## Overview

Simplifies the geostore polygon requested for countries/admins that are **absolute units**.


edbrett:
🆎 solute un 🇮🇹 's

PR #3539, opened 2018-08-07
Feature/custom legend components

edbrett:
_"They say you are what you eat, but I don't remember eating a flipping legend"_ - _new GFW legend (probably)_

The time has come to customise the WRI API legend with both styles and some fancy custom components to control our every complicated layers. Here is what we have:

- `Timeline`: a range slider timeline component that allows you to trim, filter and play canvas players on the map
- `LayerListMenu`: displays a list of togglable layers when the layer is type `isMultiLayer`, with additional info component nested inside.
- `LayerSelectorMenu`: displays a selector menu for when the layer is type `isSelectorLayer`
- `LossStatement`: a disclaimer for loss type layers
- `Threshold Selector`: a dropdown selector for layers with `thres` or `threshold` in their `params_config`

Other new components for a better life and a better you:
- `Toggle`: reimplemented the `LayerToggle` component with a new custom toggle for dynamic styling. 
- `Dropdown`: this is now allows a native params to use the native browser selector

Additional needed tasks:
- Refactor datasets parsing and layer config building throughout app
- General styles for legend and menu
- Lots of bug fixes for layers and layersConfig
- Remove moment from any date parsing inside loops


edbrett:
I cant reproduce the error with GLADs losing the timeline but that might have been a deps issue. Regarding timeline, as this is an interaction design problem I am going to merge and we can work with Fausto on the use cases later on. Thanks for the good feedback 💯 .

PR #3554, opened 2018-08-16
fixed bug on active fires analysis service

benlaken:
## Overview
Fires analysis service (map --> draw shape) is currently giving back loss instead of active fires, this PR returns the correct value.


01painadam:
Added WIP - I'll work off this branch to fix the admin issue

PR #3557, opened 2018-08-17
Feature/glads by latest

edbrett:
The GLADs layer has some unique cases that we need to handle:

1. We need to fetch the latest date of the data from the APIs latest endpoint. This allows us to show recent alerts on the map.

2. Filter layer but confirmed alerts only. This requires updating the decode function and adding a custom component to the legend to handle this.

Both of these cases have been covered and some additional fixes made to cover error in the decode from a previous PR 😭 .


app/javascript/components/map/components/legend/legend.js
```diff
@@ -136,6 +136,20 @@ class Legend extends PureComponent {
...
 
+  setConfirmed = layer => {
+    const { layers, setMapSettings } = this.props;
+    const { dataset } = layer;
+    const datasetIndex = layers.findIndex(l => l.dataset === dataset);
+    const newLayers = [...layers];
+    let newDataset = newLayers[datasetIndex];
+    newDataset = {
+      ...newDataset,
+      confirmedOnly: true
```
sorodrigo: shouldn't this be in a reducer?

app/javascript/components/map/map-decode-config.js
```diff
@@ -110,33 +110,33 @@ export default {
     }
   },
   // GLADs
-  '95b34906-d998-496b-80e7-9d8e2a70cd25': {
+  'dd5df87f-39c2-4aeb-a462-3ef969b20b66': {
```
sorodrigo: This is begging to be put in a constant somewhere

edbrett: This is the constants. We can have a separate one for mapping to the layer Id, but then we have to cross compare again inside the layer selector.

sorodrigo: nah it's fine. Didn't realised that...

PR #3558, opened 2018-08-17
Fix Admin-0 Fires Widget

01painadam:
## Overview

The admin 0 response is now too heavy and returns an error. Introduced a new service to fetch aggregated data at adm0 that uses elastic instead.


01painadam:
Added comment in code to explain reasoning behind using 2 services.

PR #3559, opened 2018-08-20
Adds IDN-specific header sentence

01painadam:
## Overview

Indonesia requires a bespoke header. This does that.

Adds:

- primary forest loss fetch
- indonesia sentence
- logic and data parsing for primary forest in IDN

APPROVED


01painadam:
That last commit fixes an issue from the feedback spreadsheet that was mentioned.

PR #3560, opened 2018-08-20
Fix/map in dashboards

edbrett:
After many updates to the new map it had broken the map in the dashboards. As we aren't certain exactly how we are going to integrate the two and what feature we want on the dashboards with the new map I have rolled the dashboards map back (now `map-old` inside the react components. This means we can build the new as we like and not have to worry about version v3 affecting production when we release the demo.

- Add `map-old`
- Integrate new Geostore provider back into old map
- Fix some small bugs in the widget settings menu

This also allows us to put the new map on staging without any risk off affecting external pages.


sorodrigo:
@01painadam I think it's better if you review this one. We don't have to re-review the code just test the application.

PR #3561, opened 2018-08-21
Removes 7-day fires widget

01painadam:
## Viking Burial

Goodbye fires. We barely knew ye.

🔥 ⚰️ 🔥  

NOTE: Should we delete the code? May be useful in the future...


01painadam:
Yeah, for now I've just disabled it in the initial config.

edbrett:
Can we remove it from the manifest also? Otherwise it gets imported on page load.

PR #3574, opened 2018-08-29
[MAP ANALYSIS] Increase polygon complexity for 'small' countries

01painadam:
## Overview

The map analysis highlights the selected country with low complexity. I have upped the complexity for all iso except 'very large' countries.

Shouldn't impact performance.

_Before:_

screen shot 2018-08-29 at 09 54 46

_After:_

screen shot 2018-08-29 at 09 56 19


edbrett:
Does this not need updating in the dashboards map also?

PR #3591, opened 2018-09-26
Feature/widgets v2

edbrett:
## Overview

In order to integrate the widgets in to the new map analysis a small refactor is needed to make them dumb to the location that are being shown for. This update each widget and:
- tidies up the top level widgets parent component that handle passing global data to each widget
- dynamically parses widget options and location by type
- adds the relevant widgets to the analysis window
- syncs widget state with map layer state and vice versa, allowing timelines, hover interactions and settings to stay in sync for both.
- add simple view mode to widgets and charts.

Improvements include:
- reduces passing of props to each widget child component
- only refetch data for each widget when location changes or settings to reduce fetching
- simplify the boiler plate code for each widget
- make more generic styles and allow styling from page


app/models/application.rb
```diff
@@ -10,9 +10,9 @@ def find_all_countries
       end
     end
 
-    def find_country_by_iso(iso)
-      return nil unless iso
-      url = "#{ENV['CARTO_API']}/sql?q=SELECT%20iso%2C%20name_engli%20as%20name%20FROM%20gadm36_countries%20WHERE%20iso%20%3D%20%27#{iso}%27%20AND%20iso%20!%3D%20%27XCA%27%20AND%20iso%20!%3D%20%27TWN%27"
+    def find_adm0_by_adm0_id(adm0)
+      return nil unless adm0
+      url = "#{ENV['CARTO_API']}/sql?q=SELECT%20iso%2C%20name_engli%20as%20name%20FROM%20gadm36_countries%20WHERE%20iso%20%3D%20%27#{adm0}%27%20AND%20iso%20!%3D%20%27XCA%27%20AND%20iso%20!%3D%20%27TWN%27"
```
simaob: if I can barge in, this could be made more readable by using the following format:

````
url = "#{ENV['CARTO_API']}/sql?q="
url += <<-SQL
  SELECT iso, name_engli as name
  FROM gadm36_countries
  WHERE iso = '#{adm0}' AND iso != 'XCA' AND iso != 'TWN'
SQL
````

Should still work!

edbrett: Ah thank you! Barge away 🛥

PR #3597, opened 2018-10-04
Update Extended GLADs

01painadam:
## Overview

Updates glad extent and coverage


app/assets/javascripts/map/views/layers/GladCoverageLayer.js
```diff
@@ -1,13 +1,13 @@
 define([
-  'abstract/layer/CartoDBLayerClass',
-], function(CartoDBLayerClass) {
+  'abstract/layer/CartoDBLayerClass'
+], (CartoDBLayerClass) => {
```
edbrett: This breaks IE. Can you revert to the old syntax?

01painadam: Mustve been linted

PR #3601, opened 2018-10-10
Feature/map search

edbrett:
## Overview

screen shot 2018-10-10 at 21 01 59

Have you ever wanted to find yourself? Find something more? Well, if yourself is either a GFW map layer OR a location that is defined within the GADM36 political boundaries then have we got the search for you! So how can you use this deep, soulful search to find you:

- Search by dataset or location and a list of layers and locations appear based on escaped (by [comma], [space] and [comma][space]). You can add these to the map by clicking them.
- Search by coordinates
- Search by decimal degrees
- Search by UTM coordinates.

BONUS fixes:
- css fixes for position of analysis loader
- tidy up map v2 page code which was doing nothing special, just free loading.


benlaken:
Testing, and it seems like no admin-0 location is selectable from the search: (I tried with India, UK, Indonesia, USA)

PR #3643, opened 2018-11-14
Feature/sync basemaps with label

blayhem:
## Overview

Labels on map needed better contrast depending on which basemap is used, so now the labels change automatically when the basemap is changed (the opposite is not true).

Label used for each map is predefined in the basemaps schema as the key `labelsKey` (e.g. 'lightLabels' or 'default', which corresponds to the key in the `labels` object exported).


app/javascript/components/map-v2/components/basemaps/basemaps-schema.js
```diff
@@ -35,6 +35,7 @@ export default {
     id: 'default',
     value: 'default',
     label: 'default',
+    labelsKey: 'default', // dark labels
```
edbrett: This comment seems valid but maybe you could move it from being a specific key?

PR #3644, opened 2018-11-15
Showing disclaimer for analysis data from Z0-Z10 only

blayhem:
## Overview

Only showing the analysis message "This algorithm approximates the results by sampling the selected area. Results are more accurate at closer zoom levels." when you are zoomed out (on zoom levels 0 to 10).

### DIFF
--- a/app/javascript/components/map-v2/components/analysis/components/draw-analysis/component.jsx
+++ b/app/javascript/components/map-v2/components/analysis/components/draw-analysis/component.jsx
@@ -79,7 +79,8 @@ class DrawAnalysis extends PureComponent {
       handleShowDownloads,
       showDownloads,
       downloadUrls,
-      hasLayers
+      hasLayers,
+      zoomLevel
     } = this.props;
 
     return (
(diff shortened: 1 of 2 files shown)


app/javascript/components/map-v2/components/analysis/components/draw-analysis/component.jsx
```diff
@@ -166,11 +167,13 @@ class DrawAnalysis extends PureComponent {
                 </ul>
                 {showWidgets && <Widgets simple analysis />}
                 <div className="disclaimers">
-                  <p>
-                    This algorithm approximates the results by sampling the
-                    selected area. Results are more accurate at closer zoom
-                    levels.
-                  </p>
+                  {zoomLevel < 11 && (
```
edbrett: Just for consistency with the rest of the components, could we call this just `zoom`?

blayhem: I thought about it, but as zoom can also be a verb (function) I ended up with zoomLevel, which I considered to be more specific. But I can change it back to zoom if you prefer it!

edbrett: I am happy with that, and you are correct it is a better name. We can leave it like this.

PR #3660, opened 2018-11-26
Feature/latest provider

edbrett:
## Overview

When loading the map for the first time we are fetching all datasets and then manually deciding which latest endpoints to find. This is slow and tricky to maintain. In the API, on the params and decode params config we thave the URL for the `endDate` key which we can use to get the latest endpoint. 

SO this PR adds a latest provider component which when the layers change on the map goes and gets any new latest endpoint needed (only the first time around) and then updates the params. Benefits:

- map will load faster first time around
- only fetching the latest endpoints we need
- easier to maintain as it is universal for each layer

## Testing

- go to the map and add and remove forest change layers (glads, forma, terrai)
- check loader show first time adding the layer
- check timeline doesnt have current date


edbrett:
@blayhem so for example you should see 16th Nov as the last possible date for GLADs? 
screen shot 2018-11-27 at 10 05 58

edbrett:
@blayhem fixed!

PR #3661, opened 2018-11-26
Correct calculations for plantations

01painadam:
## Overview

Plantations loss / extent / emissions were not being summed over all types in the returned data, instead just taking the first element in the array. Note that plantations are returned with a `bound1` or `bound2` key.

_(I suggest that we review this across all widgets that use `forestType === 'plantations'`)_

I have updated the header and loss in plantations widget to handle this. In all cases where there are plantations (and we therefor quote _natural forest_) we should have the figure in the header:

screen shot 2018-11-26 at 17 17 32
5.07kha in this case ... equal to the tooltip figure here:

screen shot 2018-11-26 at 17 17 41
natural forest: 5.07kha


app/javascript/components/widgets/widgets/forest-change/tree-loss-plantations/selectors.js
```diff
@@ -17,20 +18,33 @@ const getSentence = state => state.config && state.config.sentence;
...
+          const groupedPlantations = groupBy(lossPlantations, 'year')[d.year];
+          const summedPlatationsLoss =
+            groupedPlantations && sumBy(groupedPlantations, 'area');
+          const summedPlatationsEmissions =
+            groupedPlantations && sumBy(groupedPlantations, 'emissions');
+
+          const returnData = {
+            ...d,
+            outsideAreaLoss:
+              totalLossByYear[d.year][0].area - summedPlatationsLoss,
```
edbrett: I know this was there before but it concerns me. As it is reused a few times can we declare `totalLossByYear[d.year][0]` as a variable above and then access the keys here. Something like this:

```
const totalLossForYear = (totalLossByYear[d.year] && totalLossByYear[d.year][0]) || {};
```

app/javascript/pages/dashboards/header/header-actions.js
```diff
@@ -33,27 +33,44 @@ export const getHeaderData = createThunkAction(
...
+            const summedPlantationsEmissions =
+              groupedPlantationsLoss &&
+              sumBy(groupedPlantationsLoss[latestYear], 'emissions');
+            const summedPlantationsExtent =
+              groupedPlantationsExtent &&
+              sumBy(groupedPlantationsExtent[latestYear], 'area');
             const summedLoss = sumBy(groupedLoss[latestYear], 'area');
             const summedEmissions = sumBy(groupedLoss[latestYear], 'emissions');
             const data = {
               totalArea: (extent[0] && extent[0].total_area) || 0,
```
edbrett: Again here we aren't checking if extent is an array or even if it exists. Can we improve this?

app/javascript/components/widgets/widgets/forest-change/tree-loss-plantations/actions.js
```diff
@@ -10,11 +10,12 @@ export default ({ params }) =>
     .then(
       axios.spread((plantationsloss, gadmLoss) => {
         let data = {};
-        const loss = plantationsloss.data && plantationsloss.data.data;
+        const lossPlantations =
+          plantationsloss.data && plantationsloss.data.data;
         const totalLoss = gadmLoss.data && gadmLoss.data.data;
-        if (loss.length && totalLoss.length) {
+        if (lossPlantations.length && totalLoss.length) {
```
edbrett: if `lossPlantations` is undefined then this will crash also. You need to check for it first before doing an array method.

app/javascript/pages/dashboards/header/header-actions.js
edbrett: This still needs addressing. What is the extent isnt an array? This will crash.

app/javascript/components/widgets/widgets/forest-change/tree-loss-plantations/actions.js
01painadam: `if (lossPlantations && totalLoss && lossPlantations.length && totalLoss.length) {` ?

app/javascript/pages/dashboards/header/header-actions.js
01painadam: plantationsExtent is always a length = 1 array so i've gone with:

`plantationsExtent && plantationsExtent.length ? plantationsExtent[0].area : 0,`

PR #3664, opened 2018-11-27
Feature: biomass loss widget + woody biomass analysis

blayhem:
## Overview

Updates the old Climate emissions widget to add the following:
- new colours to match the associated layer
- show widget in map when layer active
- allow widget to be viewable for wdpas, geostore, and use cases

Additional features:
- Reduce analysis fetches by filtering endpoints by active widgets
- updates widgets to have default labels for wdpa, geostore, use


blayhem:
> @benlaken you can now see the widget inside the map when analysing the area, so I would say this PR is complete. Can you review?

We're having some bugs when trying to analyze a geometry or polygon for this layer (in the map). Seems like the widget never renders - just infinite spinner.

screenshot 2018-11-28 at 16 02 17
screenshot 2018-11-28 at 15 59 07

edbrett:
@benlaken @blayhem updated! But I think the endpoint is failing to regularly for this to be usable.

benlaken:
@edbrett  this lgtm - true the analysis fails often, but that is mostly out of our hands - I can possibly tweak the params of the micro-service to make it less accurate but more reliable, and try and find a balance. Seems to work well apart from that.
We could also improve the dynamic sentences in the analysis, but I think that is something that is a more general issue.
I suggest we go ahead and merge this.
screenshot 2018-11-30 at 13 08 20

PR #3671, opened 2018-12-11
Feature/module maps

edbrett:
## Overview

Currently we have a single monolith of a map that is rather over encumbered. As we will be needed the same map in the dashboards and potentially other parts of the site it would be nice if we split the map out into a simple map component and a use that to build specific layout components for other maps. 

This PR addresses this by creating a maps folder with map component inside that handles to simple component. An additional map component exist for the main map.

WARNING: this PR is pretty un-reviewable due to the amount of file index changes. I recommend checking out the branch and testing the map features themselves.

Additional fixes:
- Order of recent imagery has been fixed below the data layers
- Refetching of analysis when no params or location change has been made has been fixed
- an old bug with import reducers from recent imagery has been fixed


benlaken:
LGTM except for links to view on map on the Explore --> topics cards.

PR #3677, opened 2018-12-19
Feature/whrc biomass widget (simple sentence)

blayhem:
## Overview

This widget features a ranked list and information about biomass density and total biomass of a region. It has settings to change tree cover threshold, as well as a "view on map" button (above ground biomass density).

On global, it shows the top 5 countries by total biomass, and on every other region the ranked list is ordered according to the variable that is selected. The list shows the current region in the middle, and the two above and below.

The widget uses the data from the whrc-biomass endpoint.

image


blayhem:
Note when merging: file ` app/javascript/services/climate.js` integrates with the same (new) file in [this](https://github.com/Vizzuality/gfw/pull/3681) PR.

app/javascript/components/numbered-list/numbered-list-component.jsx
```diff
@@ -39,7 +39,10 @@ class NumberedList extends PureComponent {
                   </div>
                   {typeof item.value === 'number' && (
                     <div className="item-value">
-                      {formatNumber({ num: item.value, unit })}
+                      {formatNumber({
+                        num: item.value,
+                        unit: item.unit ? item.unit : unit
```
edbrett: why not `item.unit || unit` ?

PR #3679, opened 2018-12-26
Feature/donut plot

blayhem:
## Overview

Pie chart: Emissions in plantations vs natural forests. Feats a donut plot showing CO2 emissions of Natural forest and Plantations.

image


benlaken:
* Title: "Emissions from biomass loss in Natural Forests vs plantations in <X>"
* We should not have a global version of this widget, as it needs to be limited to countries with plantations data. (i.e. same restrictions as the widget 'Forest loss in Natural Forest in <X>'.)
* The first year shown should be 2013 (same as the other plantation widget previously mentioned). The last year should be the last year of biomass loss data returned - 2017 currently. (Currently the dynamic sentence states the data covers a period of 2010 to 2018.)
* Years should be selectable in the settings menu (see example 'Emissions from biomass loss in <x>').
* Units should be selectable in settings menu, changing between CO₂ and Carbon. 
--> check widget 'Emissions from Biomass Loss in <X>': to see how to change between CO₂ and Carbon units.
* CO₂ should appear as sub-script (₂), not as CO2.
* Canopy density should be selectable 
* Units on widget legend need to be changed: they are currently in ha instead of tonnes (t). Note, the units in the dynamic sentence are correct.
* Associated dataset layer should be biomass loss (not tree cover). The layer ID for this is: `b32a2f15-25e8-4ecc-98e0-68782ab1c0fe`
* Associated metadata key for biomass loss data is `tree_biomass_loss`
* Settings button currently not working

PR #3687, opened 2019-01-15
Update landing_controller.rb

simaob:
## Overview

Added a bit extra detail on one of the descriptions on the homepage.


simaob:
Served its purpose.

PR #3701, opened 2019-02-01
Feature/new footer

blayhem:
## Overview

This PR features a "new" footer for the entire site, intended to be the same as the old one but implemented in React, as well as 2 different modals ('Contact us' and 'Newsletter').


app/javascript/components/footer/footer-component.jsx
```diff
@@ -0,0 +1,207 @@
...
+import youtube from 'assets/icons/social/youtube.svg';
+import partners from 'data/partners.json';
+import './footer-styles.scss';
+
+class Footer extends PureComponent {
+  render() {
+    return (
+      <MediaQuery minWidth={850}>
+        {isDesktop => (
+          <div className="footerGfw">
```
edbrett: So whilst we are doing this lets update the code to be inline with our other components. Lets start using the same class structure etc. so rename this to `c-footer` or similar.

app/javascript/components/modals/contact-us/styles.scss
```diff
@@ -0,0 +1,87 @@
+@import '~styles/settings.scss';
+$mobile-padding: rem(20px);
+$desktop-padding: rem(40px);
+
+.c-form-container {
```
edbrett: Can we use better naming for these across all the modals?

app/javascript/components/modals/contact-us/styles.scss
```diff
@@ -0,0 +1,87 @@
...
+  .subtitle {
+    margin-top: 0.3125rem;
+    font-size: 0.8125rem;
+    line-height: 1.2;
+    color: #999;
+    margin-bottom: 1.875rem;
+  }
+}
+
+.feedback-message {
```
edbrett: This isn't scope to the class of the component. If another class is created with the same name it will be polluted.

app/javascript/components/modals/newsletter/component.jsx
```diff
@@ -0,0 +1,52 @@
...
+        {isDesktop => (
+          <Modal
+            isOpen={open}
+            contentLabel="Contact Us"
+            onRequestClose={() => {
+              setModalNewsletterOpen(false);
+            }}
+          >
+            {
+              <div className="c-form-container">
```
edbrett: Better naming here.

app/javascript/components/modals/newsletter/styles.scss
```diff
@@ -0,0 +1,90 @@
+@import '~styles/settings.scss';
+$mobile-padding: rem(20px);
+$desktop-padding: rem(40px);
+
+.c-form-container {
```
edbrett: And here.

app/javascript/utils/stateToUrl.js
```diff
@@ -28,7 +28,12 @@ export const encodeStateForUrl = params => {
 export const setComponentStateToUrl = ({ key, subKey, change, state }) => {
   const { location } = state();
   let params = change;
-  if (location.query && location.query[subKey || key]) {
+  if (
```
edbrett: 😄

app/javascript/components/modals/contact-us/styles.scss
```diff
@@ -2,86 +2,89 @@
...
-    margin-bottom: 30px;
-  }
-
-  h3 {
-    font-weight: 500;
-    font-size: 1rem;
-    text-transform: uppercase;
-    color: #555;
+.c-contact-us {
+  .c-contact-form {
```
edbrett: Another thing we try to avoid is referencing component parent classes inside a css file. So in this case you are changing the components styles from the outside, but if there were multiple instances of this component rendered and you only wanted to change this instance you would have a problem. Can you try to assign a theming class here to adjust layout as desired.

app/javascript/components/modals/contact-us/styles.scss
```diff
@@ -2,86 +2,89 @@
...
-  display: flex;
-  justify-content: center;
-  align-items: flex-start;
-  flex-direction: column;
-  background-color: #fff;
-  max-width: 70rem;
-  width: 600px;
-  padding: 1rem;
-  margin: auto;
+    .c-button {
```
edbrett: same here.

PR #3705, opened 2019-02-14
Feature/metadata modal url

blayhem:
## Overview

This PR changes the metadata modal triggering behavior to bind the metadata key to the URL so that it can be sharable.


edbrett:
When testing I have found the canopy density info button is no longer working on the widgets settings menus, and the same for the info buttons in the land category or forest type dropdowns.

PR #3706, opened 2019-02-19
Feature/soil organic carbon

blayhem:
## Overview

This PR features a new widget very similar to the whrc one: a ranked list for soil organic carbon.

image


app/javascript/components/widgets/widgets/climate/soil-organic/config.js
```diff
@@ -0,0 +1,21 @@
+export default {
+  widget: 'soil-organic',
```
edbrett: These have normally been camel case right?

PR #3707, opened 2019-02-20
Fix: Ranked gain widget parent location

01painadam:
## Overview

The ranked gain widget should have access the parent locations (e.g. Acre's parent is Brazil, Brazils is Global etc)

This update the 'getLocationsObject' function to include parent information as well as teh admin level of the selected location.

screen shot 2019-02-20 at 11 53 58

screen shot 2019-02-20 at 11 54 59


01painadam:
It's the `'<0.1%'` bug again. I've fixed it for all widgets.

PR #3708, opened 2019-02-20
Feature/landsat tiles refactor

blayhem:
## Overview

This PR uses the new landsat endpoint to fetch the tiles instead of manually calculating it for zoom values greater than 11.

### DIFF
--- a/app/javascript/components/maps/map/index.js
+++ b/app/javascript/components/maps/map/index.js
@@ -18,8 +18,7 @@ const actions = {
 class MapContainer extends PureComponent {
   static propTypes = {
     basemap: PropTypes.object,
-    mapOptions: PropTypes.object,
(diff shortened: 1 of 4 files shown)


app/javascript/components/maps/map/index.js
```diff
@@ -34,16 +33,21 @@ class MapContainer extends PureComponent {
...
+      layerBbox
     } = this.props;
 
     // update landsat basemap when changing zoom
-    if (basemap.value === 'landsat' && zoom !== prevProps.zoom) {
-      setLandsatBasemap({
-        basemap,
-        year: basemap.year,
-        zoom
+    if (
```
edbrett: If the zoom is now part of the url and is fetched from the service I don't think this is still needed?

PR #3722, opened 2019-03-15
Feature: cookies banner

blayhem:
## Overview

This PR features a WIP cookies banner (currently displaying only on the G&F page), using `localStorage` in the same way as the map helper.


app/javascript/components/cookies/index.js
```diff
@@ -0,0 +1,10 @@
+import { connect } from 'react-redux';
+import Component from './component';
+
+const mapStateToProps = ({ modalContactus, modalNewsletter }) => ({
```
edbrett: This is copied and not needed.

app/javascript/components/cookies/component.jsx
```diff
@@ -0,0 +1,42 @@
...
+
+import './styles.scss';
+
+class Cookies extends PureComponent {
+  render() {
+    const { agreeCookies } = this.props;
+    return (
+      <div className="c-cookies">
+        <div className="row">
+          <div className="medium-1" />
```
edbrett: This wont work correctly without the column class.

PR #3737, opened 2019-04-02
Feature/refetch button

blayhem:
## Overview

This PR adds a Try Again button to the widgets and the recent imagery settings to force a refetch of the data in case the previous one had failed.


app/javascript/components/maps/main-map/components/recent-imagery/components/recent-imagery-settings/recent-imagery-settings-styles.scss
```diff
@@ -103,6 +103,21 @@
     }
   }
 
+  .refresh .message {
```
edbrett: The only improvement I can think of would be to make this a ui component as you have the same styles and button in two places?

PR #3750, opened 2019-04-23
Feature/ui kit map settings menu

blayhem:
## Overview

This PR features the new redesign of the settings menu in the map, as seen in the GFW mobile map designs and the basecamp thread for the new map settings.


edbrett:
I merged develop as there were some functional changes with how basemaps are now selected (which as a result had changes with the styles). I think some things might have broken slightly in the merge with paddings. Can you review again?

PR #3761, opened 2019-05-07
Feature/sandbox tab

01painadam:
## Overview

Sandbox tab in staging for experimental layers (using Vocabulary with `categoryTab = ['sandbox']`)

## Feature To Do

- Better icon? ¯\\_(ツ)_/¯ 
- Logic to pull all datasets with `categoryTab = ['sandbox']` or `categoryTab = ['test']` and ignore all other items in the vocab list


edbrett:
Doesn't seem to work. Have you created any layers? It needs at least one to render.

PR #3766, opened 2019-05-16
Fixed percentage numbers over 100

blayhem:
## Overview

Fixes incorrect % for countries in tree cover loss list widget, which sometimes were more than 100%.


app/javascript/components/widgets/widgets/forest-change/tree-loss-ranked/selectors.js
```diff
@@ -108,17 +108,26 @@ export const parseData = createSelector(
...
-        query,
-        id: d.id
-      }),
-      value: settings.unit === 'ha' ? d.loss : d.percentage
-    }));
+    return dataTrimmed.map(d => {
+      let value;
+      if (settings.unit === 'ha') {
+        value = d.loss;
+      } else if (d.percentage <= 100) {
```
edbrett: This doesn't fix the issue of the percentages being greater than 100, it just limits them. We need to make sure the percentage calculation is correct in the first place.

PR #3771, opened 2019-05-21
Fixing analytics events through the page

blayhem:
## Overview

This PR fixes some analytics events that are not being triggered (WIP).

Currently it fixes the analytics events in the dashboards, but map and homepage events are pending.

### DIFF
--- a/app/javascript/components/maps/map/components/draw/index.js
+++ b/app/javascript/components/maps/map/components/draw/index.js
@@ -4,7 +4,6 @@ import PropTypes from 'prop-types';
 import isEqual from 'lodash/isEqual';
 import reducerRegistry from 'app/registry';
 import MapboxDraw from '@mapbox/mapbox-gl-draw';
-import { track } from 'app/analytics';
 
 import '@mapbox/mapbox-gl-draw/dist/mapbox-gl-draw.css';
 import './styles.scss';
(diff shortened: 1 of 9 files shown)


app/javascript/components/maps/map/components/draw/index.js
```diff
@@ -59,6 +61,7 @@ class MapDraw extends PureComponent {
 
   closeDrawing = () => {
     const { map } = this.props;
+    track('analysisDrawCancel');
```
edbrett: Only thought here is that this will happen when a draw is completed due to closing the shape and not only when you cancel. Maybe this isnt a big problem, as we can compare to real 'drawComplete' events and see the difference.

app/javascript/components/maps/map/index.js
```diff
@@ -203,9 +203,15 @@ class MapContainer extends PureComponent {
   }, 300);
 
   handleMapInteraction = e => {
-    const { draw, menuSection } = this.props;
+    const { draw, menuSection, selectedInteraction } = this.props;
     if (!draw && !menuSection && e.features && e.features.length) {
       this.props.setInteraction(e);
+      track('mapInteraction', {
+        label:
+          selectedInteraction &&
```
edbrett: This will be the selected interaction of the previous click. All the current event data is inside e. Update to reflect.

app/javascript/components/maps/map/components/draw/index.js
blayhem: Removed, it's already implemented [here](https://github.com/Vizzuality/gfw/blob/d8baa147b66268aadcd4658561ace0ac94ea89a3/app/javascript/components/maps/components/analysis/components/chose-analysis/component.jsx#L103).

PR #3791, opened 2019-06-17
Feature/land categories and forest types on widgets map

blayhem:
## Overview

This PR adds the land category and forest type layers to the widgets' map when selecting those in the widget settings.


app/javascript/data/forest-types.json
```diff
@@ -3,24 +3,56 @@
...
     "label": "Mangrove forests",
     "value": "mangroves",
     "metaKey": "mangrove_2010_gmw",
-    "global": true
+    "global": true,
+    "datasets": [
+      {
+        "dataset": "b3b8ca9d-a071-4383-b898-d7f64573b51f",
+        "layers": [
+          "49e3f955-2bab-4f73-8589-c9c039b8da29",
```
edbrett: this is a tricky one. Each mangrove layer here is for a specific year and by category too. We should only show one layer unless we give them control over which layer is displaying. And ideally the layer that is used to intersect with the data. As the key above specifies that 2010 is the year they are using, let only put the 2010 layer.

app/javascript/components/widgets/components/widget/component.jsx
```diff
@@ -87,6 +94,22 @@ class Widget extends PureComponent {
...
+          .map(
+            polyname =>
+              polyname.datasets &&
+              polyname.datasets.map(d => ({
+                opacity: 0.7,
+                visibility: 1,
+                ...d
+              }))
+          )
+          .flat(1)) ||
```
edbrett: You could also use flatMap here too!

PR #3800, opened 2019-07-01
Feature/dynamic sentence translation

edbrett:
## Overview

Add `notranslate` class to the dynamic sentence component so that transifex ignores the variables and allows the translation of the widget sentences.

## Testing

You need access to transifex. Go to `0.0.0.0:5000?transifex` and login. Then inspect the widget sentences using the transifex live interface.

### DIFF
--- a/app.json
+++ b/app.json
@@ -158,6 +158,9 @@
     "SPARKPOST_SMTP_USERNAME": {
       "required": true
     },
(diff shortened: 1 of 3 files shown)


edbrett:
@blayhem we will have to review this one together as you don't have access.

PR #3804, opened 2019-07-09
Fixed empty values that crashed the settings menu

blayhem:
## Overview

When retrieving planet basemap data, the default selected value caused a conflict with the empty data retrieved, causing the basemap settings menu to crash. Added a couple of checks to avoid this.

image


edbrett:
I am still getting issues with quarterly not showing up.

PR #3823, opened 2019-07-31
Feature/vizz components timeline

edbrett:
## Overview

An upgrade to the vizz components library unifying the timeline and slider components across the app

- remove timeline and slider components from GFW
- add timeline and slider components from lib to GFW
- change css theme to `vizzuality` from `wri-api`

## Notes

The v3 of vizzuality-components is a big update with breaking changes to the legend components such as toolbar and slider. This shouldn't affect GFW as we weren't using them before, but these components originally came from GFW so 🤷‍♂ let's test properly. 

## Testing

- add GLADs and tree cover loss and play the timelines on the map
- use recent imagery slider to cloud coverage


app/javascript/components/map/components/legend/components/timeline/index.js
```diff
@@ -34,184 +24,36 @@ const mapStateToProps = (
...
-      min: 0,
-      max: dateDiffInDays(maxDate, minDate),
-      start: dateDiffInDays(startDate, minDate),
-      end: dateDiffInDays(endDate, minDate),
-      trim: dateDiffInDays(trimEndDate, minDate),
-      loops: 0
-    };
-  }
-
-  componentDidUpdate(prevProps, prevState) {
```
blayhem: I guess most of this logic is now in vizzuality-components?

edbrett: yup, nearly all of it!

PR #3825, opened 2019-08-07
National Land Cover Database (USLC) widget

blayhem:
## Overview

WIP sankey widget

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,6 +115,7 @@
     "coffee-loader": "0.7.3",
     "coffee-script": "1.12.7",
     "constants": "0.0.2",
+    "cw-components": "github:ClimateWatch-Vizzuality/climate-watch-components#1.51.0",
     "d3": "4.11.0",
     "d3-format": "1.2.1",
     "deck.gl": "^7.1.10",
--- a/app/javascript/components/widgets/components/widget-sankey/component.jsx
+++ b/app/javascript/components/widgets/components/widget-sankey/component.jsx
@@ -1,0 +1,114 @@
+import React, { PureComponent } from 'react';
+// import PropTypes from 'prop-types';
+// import { format } from 'd3-format';
+
+import { Sankey } from 'cw-components';
+
+import './styles';
+
+class WidgetTreeCover extends PureComponent {
+  render() {
+    // const { data, settings, simple } = this.props;
+
+    const testData = {
+      nodes: [
+        {
+          name: 'Germany',
+          color: '#00955f'
+        },
+        {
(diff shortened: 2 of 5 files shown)


01painadam:
One small thing - if possible - could we make the default option `Changes only` appear on the left side of the toggle?

Screen Shot 2019-09-10 at 16 11 33

app/javascript/components/widgets/components/widget/components/widget-header/components/widget-settings/widget-settings-component.jsx
```diff
@@ -17,16 +17,17 @@ class WidgetSettings extends PureComponent {
...
   getUnitVariable = (items, widget, settings, onSettingsChange, type) => {
+    const unit = type === 'categorization system' ? 'source' : type;
     if (items.length <= 1) return null;
     if (items.length === 2) {
       return (
         <Switch
           theme="theme-switch-light"
-          label={type === 'unit' ? 'UNIT' : 'VARIABLE'}
-          value={settings[type]}
+          label={String(type).toUpperCase()}
```
edbrett: Why `String()`?

app/javascript/utils/strings.js
```diff
@@ -13,3 +15,50 @@ export const getIndicator = (forestType, landCategory) => {
   }
   return 'admin';
 };
+
+// formats/create ellipsis nodes text using # of lines available, node width, etc
+export const splitSVGText = (
+  text,
+  height,
+  _labelCharHeight,
```
edbrett: this nomenclature does seem very reusable friendly? Should it be a global util?

package.json
```diff
@@ -115,6 +115,7 @@
     "coffee-loader": "0.7.3",
     "coffee-script": "1.12.7",
     "constants": "0.0.2",
+    "cw-components": "github:ClimateWatch-Vizzuality/climate-watch-components#1.51.0",
```
edbrett: If we are using a custom sankey do we need this lib?

app/javascript/components/widgets/components/widget/components/widget-header/components/widget-settings/widget-settings-component.jsx
blayhem: To make sure the type doesn't crash the `toUpperCase`, just a quick decision until we implement the new settings.

app/javascript/utils/strings.js
blayhem: I agree, but it has to do with text. I thought this could be a good place, but maybe some other `utils` file would be more appropriate?

package.json
blayhem: You're right, removed 👍

PR #3828, opened 2019-08-16
Feature/public areas

edbrett:
## Overview

Allows users who are not logged in to view areas of interest from a share link.
- public share view for map area and dashboard
- area meta data hidden unless user logged in viewing own area
- change geodescriber to fetch based on language
- allows fetching and filtering of areas through selectors shared in areas provider
- remove subscribe modal
- fixes show, share and dashboard buttons in analysis

## Testing

- View the following area id: `598825ca6e68340010b582cd`, and you should see dashboards and map views without tags or meta data
- View you own areas as before


app/javascript/pages/dashboards/components/header/selectors.js
```diff
@@ -123,19 +134,37 @@ export const getSelectorMeta = createSelector([selectLocation], location => {
...
+      return 'save to my gfw';
+    }
+
+    return 'share dashboard';
+  }
+);
+
+export const getErrorMsg = createSelector(
+  selectError,
+  error => (error === 401 ? 'Area is private' : 'Area not found')
```
blayhem: nice one here

app/javascript/services/forest-data.js
```diff
@@ -178,7 +186,7 @@ export const getExtentGrouped = ({
     .replace('{extentYear}', getExtentYear(extentYear))
     .replace('{WHERE}', getWHEREQuery({ iso: adm0, adm1, adm2, ...params }));
 
-  return request.get(url);
+  return axios.get(url, { cancelToken: token });
```
blayhem: what is the cancelToken for?

edbrett: if you try to fetch again for some reason, the first fetch gets cancelled.

PR #3829, opened 2019-08-16
Feature/areas widgets

edbrett:
## Overview

Screenshot 2019-08-16 at 16 27 13

Adds the tree cover loss widget to all areas regardless of geostore or path.


blayhem:
Maybe we can get the area name in the widget to display that instead of "your custom area"? Would that make sense?

edbrett:
Yes, that is my next set of tasks. Metadata 🎉

edbrett:
imgt to keep it small.

PR #3838, opened 2019-08-28
Update fires to use most recent data

01painadam:
## Overview

Uses the data to find the latest week and year value.

Screen Shot 2019-08-28 at 16 05 32

## Testing

Please test with prod table and dataset. Should get to w/o 19th Aug.

## NOTES

**WORKS ASSUMING WE HAVE WEEKS WITH ZERO DATA**

### DIFF
--- a/app/javascript/components/widgets/widgets/forest-change/fires-alerts/selectors.js
+++ b/app/javascript/components/widgets/widgets/forest-change/fires-alerts/selectors.js
@@ -1,5 +1,6 @@
 import { createSelector, createStructuredSelector } from 'reselect';
 import isEmpty from 'lodash/isEmpty';
+import maxBy from 'lodash/maxBy';
 import { format } from 'd3-format';
 import groupBy from 'lodash/groupBy';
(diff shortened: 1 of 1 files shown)


app/javascript/components/widgets/widgets/forest-change/fires-alerts/selectors.js
```diff
@@ -24,12 +25,22 @@ export const getData = createSelector(
...
     const years = [];
-    const latestFullWeek = moment().subtract(2, 'w');
     const lastWeek = {
-      isoWeek: latestFullWeek.isoWeek(),
-      year: latestFullWeek.year()
+      isoWeek:
+        dataMaxWeek ||
+        moment()
+          .subtract(2, 'w')
+          .week(),
```
edbrett: don't we want isoWeek here like above removed code?

01painadam: No, should be isoWeek. Fixed.

PR #3850, opened 2019-10-01
Updated contact us form

blayhem:
## Overview

image

Changed the contact us form with the changes WRI requested, plus some fixes and new functionality for the `select` Field in form-fields: it admits now a placeholder that adds an extra option, and adds a validator to use when the select needs a valid value.

It's WIP because it's probably missing work on the mail micro service.


edbrett:
In general most of the items from the basecamp list aren't complete. The topic still has a field selected by default, there is no option for 'not applicable' for the tools fields, a full stop in the title. Please check more carefully before requesting review.

PR #3865, opened 2019-12-12
Feature/aoi webhook links

blayhem:
## Overview

"Preview of payload" and "Test webhook" buttons now work as intended.

TODO: maybe? Improve readability on the webhook test errors, validation, etc.

image


app/javascript/components/modals/save-aoi/components/save-aoi-form/index.js
```diff
@@ -1,3 +1,6 @@
+import { connect } from 'react-redux';
+import * as modalSourcesActions from 'components/modals/sources/actions';
 import Component from './component';
 
-export default Component;
+// export default Component;
```
edbrett: tidy up

PR #3870, opened 2019-12-18
Fixed lower caps for Key Biodiversity Areas indicator

blayhem:
## Overview

Related [thread](https://basecamp.com/3063126/projects/10727890/todos/404782334).

This PR fixes the capitalization of Key Biodiversity Areas in the dashboard widgets. However, because it's the only indicator that is kept with the original capitalization (see thread), I had to add a lot of "special cases" in the widget configs, and there's probably a better way of doing this.

### DIFF
--- a/app/javascript/components/widgets/components/widget/components/widget-footer/selectors.js
+++ b/app/javascript/components/widgets/components/widget/components/widget-footer/selectors.js
@@ -61,7 +61,10 @@ export const getStatements = createSelector(
         ? translateText(
           '*{indicator} are available in {datasetsCount} countries only',
           {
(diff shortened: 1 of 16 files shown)


blayhem:
> Ok, I understand your point. Its tricky to place this, but I don't think we can have all these exceptions happening on each widget. Can you try to change this to be at the widget selector level? I think you could do it here and remove the .toLowerCase() at the individual widget level? https://github.com/Vizzuality/gfw/blob/develop/app/javascript/components/widgets/components/widget/selectors.js#L181

Thank you for the suggestion, this is a much cleaner approach. Done 👍

PR #3886, opened 2020-01-23
Feature/climate modal

blayhem:
## Overview

GFW modal to inform GFW climate users. Try it [here](https://gfw-staging-app-pr-3886.herokuapp.com/?gfwclimate=true).

image

### DIFF
--- a/app/javascript/components/modals/climate/styles.scss
+++ b/app/javascript/components/modals/climate/styles.scss
@@ -10,21 +10,20 @@ $mobile-margin: rem(10px);
     margin-top: auto !important;
   }
 
-  .contact-us-content {
-    .feedback-message {
-      display: flex;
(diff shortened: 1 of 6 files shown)


edbrett:
How do I view it?

blayhem:
> How do I view it?

I updated the description, but [here](https://gfw-staging-app-pr-3886.herokuapp.com/?gfwclimate=true)

app/javascript/components/modals/climate/styles.scss
```diff
@@ -0,0 +1,30 @@
+@import '~styles/settings.scss';
+$mobile-padding: rem(20px);
+$desktop-padding: rem(40px);
```
edbrett: Do we need these declared here if they are never used?

PR #3895, opened 2020-02-03
[Feature - AOIs] AOIs profile form

blayhem:
## Overview

WIP Profile modal to update profile data from myGFW. Introduces new fields such as firstName, lastName, topics interested in, aoiLocation, etc.

image


app/javascript/components/forms/components/field-wrapper/styles.scss
```diff
@@ -2,18 +2,10 @@
 
 .c-form-field {
   font-size: rem(14px);
-  margin-bottom: rem(25px);
+  margin: rem(12px) 0;
```
edbrett: Why this change?

app/javascript/components/forms/components/field-wrapper/styles.scss
```diff
@@ -57,6 +49,14 @@
     transition: all 0.1s ease-in-out;
   }
 
+  details {
```
edbrett: Error with the linter?

app/javascript/components/forms/profile/component.jsx
```diff
@@ -78,47 +81,74 @@ class ProfileForm extends PureComponent {
...
+                        placeholder="Select a sector"
                       />
+                      {values.sector &&
+                        sectors[values.sector] && (
+                        <Radio
+                          name="subsector"
+                          label="Role"
+                          options={sectors[values.sector].map(s => ({
+                            label: s,
+                            value: s.replace(/( )+|(\/)+/g, '_')
```
edbrett: Not sure this is totally necessary for this native select.

edbrett: In fact maybe we should refactor it to only accept arrays or string and scrap the objects all together.

app/javascript/components/forms/components/radio/component.jsx
```diff
@@ -54,6 +64,17 @@ class Radio extends PureComponent {
                       <label className="radio-label" htmlFor={id}>
                         <span />
                         {option.label}
+                        {otherRegex.test(option.label) &&
+                          selectedOption &&
+                          otherRegex.test(selectedOption) && (
+                          <Field
```
edbrett: Im not a big fan of including this inside this component. It feels like a very specific thing to the profile form. Surely just a conditional field element below in the form is simple enough? Also the styles of the input you have included are not consistent with any other input field. It should be below and the same height etc.

Screenshot 2020-02-04 at 16 15 23

app/javascript/components/forms/components/field-wrapper/styles.scss
blayhem: To add margin top to the form fields, so that they have separation on both sides (for example, to the h4 titles).

app/javascript/components/forms/components/field-wrapper/styles.scss
blayhem: yes  🤷‍♂   same functionality

app/javascript/components/forms/profile/component.jsx
blayhem: yeah I agree we should do this in the component, as it's repeated a couple of times, but I wasn't sure if we were gonna keep that after the redesign

edbrett: Lets update it then. Redesign?

app/javascript/components/forms/components/radio/component.jsx
blayhem: it's half the height on purpose, otherwise it moves the "other" option too far from the others. It was a decision not very consistent with the UI kit, but in order to respect the parent component first. 

It could be a conditional field, yes, passed as a prop. I did this for any fields having "other" and "write" keywords, as it sounds like a common case for radio buttons, but it could be an extra field in the option object, agree (less specific that way).

edbrett: In this case we need to ask the designers.

app/javascript/components/forms/components/field-wrapper/styles.scss
edbrett: Ok. For me the h4 titles are something custom to your individual form so should manage their layout and styles personally. Our form fields should be put components that always do the same things and make forms consistent. Also we don't have a design for the form with the new fields so the h4 titles do not make sense just yet. Please discuss with a designer.

app/javascript/components/forms/profile/component.jsx
```diff
@@ -61,15 +63,16 @@ class ProfileForm extends PureComponent {
...
+                    <div className="column small-12 medium-12">
                       <h1>Your profile information</h1>
                       <h3>
                         Help us help you! Tell us who you are and how you use
                         Global Forest Watch so we can better meet your needs.
                       </h3>
                     </div>
-                    <div className="column small-12 medium-6">
-                      <Input name="fullName" label="name" required />
+                    <div className="column small-12 medium-12">
```
edbrett: Why medium-12 if we already have the small-12? They both do the same thing?

app/javascript/components/forms/profile/component.jsx
```diff
@@ -61,15 +63,16 @@ class ProfileForm extends PureComponent {
                   </div>
                 ) : (
                   <Fragment>
-                    <div className="column small-12">
+                    <div className="column small-12 medium-12">
```
edbrett: Same here

app/javascript/components/forms/profile/styles.scss
```diff
@@ -9,7 +9,7 @@
 
   h3 {
     font-size: rem(16px);
-    margin-bottom: rem(40px);
+    margin-bottom: rem(28px);
```
edbrett: Reason for change?

blayhem: The top margin I added to the fields. Rolling back.

edbrett:
One thing I noticed when reviewing the staging app, let's set a max width for input fields. I haven't set this  inside the form components as in some cases we use a grid and that would break the flow, so I have been doing it on a form by form basis. Here is how I did it for contact us: https://github.com/Vizzuality/gfw/blob/develop/app/javascript/components/forms/contact/styles.scss#L18

aabdaab:
checking

app/javascript/components/forms/profile/component.jsx
```diff
@@ -77,55 +86,108 @@ class ProfileForm extends PureComponent {
...
-                        options={sectors.map(s => ({ label: s, value: s }))}
-                        placeholder="Select a language"
-                        required
+                        options={Object.keys(sectors).map(s => ({
+                          label: s,
+                          value: s
+                        }))}
+                        placeholder="Select a sector"
+                        required={
+                          source === 'AreaOfInterestModal' ? true : undefined
```
aabdaab, line 98: never set a variable to `undefined` by yourself - this is reserved to javascript engine - because when debugging you cannot tell if the value is undefined because you never defined value and js engine set it to undefined (all vars in js are initially set to undefined) or if you did somewhere. it can make debugging very confusing :P

aabdaab: especially here, this can be solved with `required={source === 'AreaOfInterestModal'}`, if the source is something else, it's gonna evaluate to false

app/javascript/components/modals/profile/component.jsx
```diff
@@ -18,16 +19,16 @@ class ProfileModal extends PureComponent {
   };
 
   render() {
-    const { open } = this.props;
+    const { open, filledProfile } = this.props;
 
     return (
       <Modal
-        isOpen={!!open}
+        isOpen={!!open || !filledProfile}
```
aabdaab, line 26: There's one case that isn't covered and it's the one with new users creating an account:
1. Users who create a MyGFW profile (not as part of saving an area) will be taken to the MyGFW profile modal window ✅ - that's working
2. Save button will be disabled until required fields complete, but the user can exit out of form without completing required fields - ❌ - that's not working, the user cannot exit the form.

We can handle that _without_ introducing a new variable `dismissed` or anything, it's better to rely on a single source of truth. We should rely solely on `open` variable and on the component _mount_ we can set the modal open if the profile isn't filled.
So:
`<Modal isOpen={open} ... >`
and in addition add componentDidMount lifecycle method
```
componentDidMount() {
    const { filledProfile, setProfileModalOpen } = this.props;
    if (!filledProfile) {
      setProfileModalOpen(true);
    }
  }
```

app/javascript/components/forms/profile/index.js
```diff
@@ -3,13 +3,37 @@ import { connect } from 'react-redux';
...
 import * as actions from './actions';
 
+const splitFirstName = fullName => {
+  const arr = fullName.split(' ');
+  return arr.slice(0, arr.length / 2).join(' ');
+};
+
+const splitLastName = fullName => {
+  const arr = fullName.split(' ');
+  return arr.slice(arr.length / 2, arr.length).join(' ');
```
aabdaab, line 13: or just `arr.pop()`

app/javascript/components/modals/area-of-interest/component.jsx
```diff
@@ -29,7 +30,9 @@ class AreaOfInterestModal extends PureComponent {
   };
 
   render() {
-    const { open, loading, loggedIn, canDelete, viewAfterSave } = this.props;
+    const { open, loading, userData, canDelete, viewAfterSave } = this.props;
+    const { email, fullName, lastName, loggedIn } = userData || {};
+    const isProfileFormFilled = !!email && (!!fullName || !!lastName);
```
aabdaab, line 35: you don't need to do this; `email && (fullName || lastName)` will work just fine

app/javascript/components/forms/profile/component.jsx
blayhem: So in this case required could be false as you said. I'm using undefined because it's the same thing as not passing the prop 'required'. Ugly but that's how it is. I'm doing this to let the library manage if it's required or not, but for now false would be more than enough, you're right 👍

app/javascript/components/modals/profile/component.jsx
blayhem: yes, we talked about this, and it's pending. I like your approach, better than using the state 🙂 thanks!

PR #3901, opened 2020-02-13
[Fix, not urgent] Fixed downloads for GLAD and VIIRS alerts on the map legend

blayhem:
## Overview

Updated the widget header component to show the download button _also_ for widgets in the map legend so everybody can download data for the widgets directly from there. This allows downloads for GLAD and VIIRS datasets from the map, and more.

### DIFF
--- a/app/javascript/components/analysis/components/show-analysis/selectors.js
+++ b/app/javascript/components/analysis/components/show-analysis/selectors.js
@@ -17,13 +17,8 @@ const selectData = state => state.analysis && state.analysis.data;
 const selectError = state => state.analysis && state.analysis.error;
 
 export const getDataFromLayers = createSelector(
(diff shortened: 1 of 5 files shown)


sorodrigo:
dont comment code, we have version control to do this.

aabdaab:
checking! 👀

aabdaab:
1. why not checking out from develop so develop can be in sync as well?
2. let's make a pair-code review or rather a functionality review.. I'm not entirely sure which parts should I be checking here ☝️#gfw-n00b
let's do it on Monday morning if you will

sorodrigo:
I think it did affected all other layers that aren’t GLAD and VIIRS and we said it wasn’t ready.

P.S. is there a way to remove me from the PR? 😅

PR #3909, opened 2020-02-26
Fix reset password for MyGFW

aabdaab:
## Overview

This PR fixes resetting password for MyGFW.
It sends the email to the API and the rest is handled there.
The issue was that the `resetUserPassword` was sending an empty string (incorrect destructuring issue)

## Testing

- Go to My GFW tab
- Click Forgot Password
- Put your email
- Hit OK
- Check that you got the email or check in the Network tab that you're sending the **Request Payload**
image

### DIFF
--- a/app/javascript/components/forms/login/actions.js
+++ b/app/javascript/components/forms/login/actions.js
@@ -34,11 +34,9 @@ export const registerUser = createThunkAction('sendRegisterUser', data => () =>
 
 export const resetUserPassword = createThunkAction(
   'sendResetPassword',
-  ({ data, success }) => () =>
+  data => () =>
     resetPassword(data)
-      .then(() => {
-        success();
-      })
+      .then(() => {})
       .catch(error => {
         const { errors } = error.response.data;
 


app/javascript/components/forms/login/actions.js
```diff
@@ -34,11 +34,9 @@ export const registerUser = createThunkAction('sendRegisterUser', data => () =>
 
 export const resetUserPassword = createThunkAction(
   'sendResetPassword',
-  ({ data, success }) => () =>
+  data => () =>
     resetPassword(data)
-      .then(() => {
-        success();
-      })
+      .then(() => {})
```
blayhem: do we even need this if we don't have `success`? These params come from `react-final-form` right?

aabdaab: we need that `then(() => {})` as otherwise we're not rendering the confirmation message (the request is still sent though). We have a `submitSucceeded` param coming from the Form that apparently is set under the hood by `react-final-form`

PR #3931, opened 2020-03-27
Hotfix/aoi data issues

01painadam:
## Overview

Fixes issue where the loss year widget could surface the year 0 in settings. Unfortunately Hansen data doesn't go back to the Roman Empire.

Loss widget - filters invalid years (namely, `treecover_loss_year = 0`) in sql
Gain - remove the 2010 extent year option from settings

## Testing

- Please double check me that this doesn't break Admin dashboard widgets!

### DIFF
--- a/app/javascript/components/widgets/climate/emissions-deforestation/index.js
+++ b/app/javascript/components/widgets/climate/emissions-deforestation/index.js
@@ -9,45 +9,46 @@ import { shouldQueryPrecomputedTables } from 'components/widgets/utils/helpers';
 
 import getWidgetProps from './selectors';
 
-const getDataFromAPI = params => fetchAnalysisEndpoint({
-  ...params,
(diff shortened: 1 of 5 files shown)


aabdaab:
I found a small issue on the Tree Cover Gain widget when checking an admin (region) shape - empty settings
Screenshot 2020-03-27 at 15 43 51

01painadam:
Ok, checking

01painadam:
Nice catch @aabdaab ! Looks like this is an existing issue 👉 staging: https://bit.ly/2UlC9hK

In this case the widget should still show three cover threshold options: i've implemented the fix 👍

01painadam:
Added WIP - gonna add a few other fixes on this branch

01painadam:
ok, nothing more to add - good to go!

PR #3933, opened 2020-04-01
Removed TSC from aois and GLAD week from map

blayhem:
## Overview

This PR removes two widgets from two views in the app, as requested:
- Loss By Driver widget from Custom AOI dashboards.
- Old GLAD widget (7 day widget) from the map analysis.

### DIFF
--- a/app/javascript/components/widgets/forest-change/glads/index.js
+++ b/app/javascript/components/widgets/forest-change/glads/index.js
@@ -20,12 +20,14 @@ export default {
   title: 'Deforestation alerts for the last week in {location}',
   categories: ['forest-change', 'summary'],
   admins: ['adm0', 'adm1', 'adm2'],
+  types: [],
   type: 'loss',
   colors: 'loss',
   sortOrder: {
     summary: 8,
     forestChange: 12
   },
+  visible: ['analysis'],
   chartType: 'composedChart',
   sentence:
     '{count} deforestation alerts detected in {location} in the last 7 days, compared to a weekly average of {weeklyMean} in the last year.',
--- a/app/javascript/components/widgets/forest-change/glads/index.js
+++ b/app/javascript/components/widgets/forest-change/glads/index.js
@@ -20,14 +20,12 @@ export default {
   title: 'Deforestation alerts for the last week in {location}',
   categories: ['forest-change', 'summary'],
   admins: ['adm0', 'adm1', 'adm2'],
-  types: [],
   type: 'loss',
   colors: 'loss',
   sortOrder: {
     summary: 8,
     forestChange: 12
   },
-  visible: ['analysis'],
   chartType: 'composedChart',
   sentence:
     '{count} deforestation alerts detected in {location} in the last 7 days, compared to a weekly average of {weeklyMean} in the last year.',
--- a/app/javascript/components/widgets/forest-change/glads/index.js
+++ b/app/javascript/components/widgets/forest-change/glads/index.js
@@ -20,7 +20,7 @@ export default {
   title: 'Deforestation alerts for the last week in {location}',
   categories: ['forest-change', 'summary'],
   admins: ['adm0', 'adm1', 'adm2'],
-  types: ['geostore', 'wdpa', 'use'],
+  types: [],
   type: 'loss',
   colors: 'loss',
   sortOrder: {
--- a/app/javascript/components/widgets/forest-change/tree-loss-tsc/index.js
+++ b/app/javascript/components/widgets/forest-change/tree-loss-tsc/index.js
@@ -57,8 +57,8 @@ export default {
     },
     // loss tsc
     {
-      dataset: '89755b9f-df05-4e22-a9bc-05217c8eafc8',
-      layers: ['fd05bc2c-6ade-408c-862e-7318557dd4fc']
+      dataset: 'f23a1803-ba17-4fe3-a1ad-39abfbfb56c6',
+      layers: ['04774cb7-912c-4612-bbd8-ba982d532c88']
     }
   ],
   metaKey: 'widget_tsc_drivers',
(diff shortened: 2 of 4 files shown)


aabdaab:
I haven't click around the app yet but shouldn't the `visible` key be changed rather than `types`?

blayhem:
> I haven't clicked around the app yet but shouldn't the `visible` key be changed rather than `types`?

for the GLADs one yes, I could remove that part too. I think the result would be the same, but if it's more legible this way I'll do it 👍

app/javascript/components/widgets/forest-change/glads/index.js
```diff
@@ -20,14 +20,12 @@ export default {
   title: 'Deforestation alerts for the last week in {location}',
   categories: ['forest-change', 'summary'],
   admins: ['adm0', 'adm1', 'adm2'],
-  types: ['geostore', 'wdpa', 'use'],
```
aabdaab, line 23: I'd bring back the types; the client's request is not about the types per-se but rather the visibility of that widget on the map. The only necessary change here is setting `visible: []` - that way if they change their mind we only need to change the visibility manifest.

app/javascript/components/widgets/forest-change/tree-loss-tsc/index.js
```diff
@@ -20,7 +20,7 @@ export default {
     initial: 'Annual tree cover loss by dominant driver in {location}',
     global: 'Global annual tree cover loss by dominant driver'
   },
-  types: ['global', 'country', 'geostore'],
```
aabdaab, line 23: I can't verify my assumption because even with the `geostore` present in the array I can't find a custom shape on the main map that would give me a different result from `0 ha`. Please, correct me if I'm wrong but I think that removing the `geostore` from the types array would mess up the analysis in the main map and we only want to hide this widget on the dashboard. So, we'd need to hide that widget only for custom AOIs and only on the dashboards.

app/javascript/components/widgets/forest-change/glads/index.js
edbrett: Agreed here. You just want to leave an empty array if you wish to keep the widget. Leave the visible array as if that is false then a widget always renders. The combo of a valid location, whitelist, and visible array is what allows the widget to be rendered.

app/javascript/components/widgets/forest-change/glads/index.js
```diff
@@ -20,14 +20,12 @@ export default {
...
   type: 'loss',
   colors: 'loss',
   sortOrder: {
     summary: 8,
     forestChange: 12
   },
-  visible: ['analysis'],
```
edbrett, line 30: Keep this as this change doesn't affect visibility of widget

app/javascript/components/widgets/forest-change/tree-loss-tsc/index.js
edbrett: So, the map layer for this widget is what is driving the need for conditional rendering. The map layers resolution begins to become invalid below zoom level 4. Activate the layer on the map and try zooming in past 4 and you'll notice that the pixels of the layer don't resample. 

If we are disabling this widget for custom AOIs one solution might be to introduce a custom `shouldRender` or similar prop that you can attach to a widget that receives all props after all other filters and you can do what ever you want. Run some turf size function to decide if it should be hidden etc. Somewhere here would work: https://github.com/Vizzuality/gfw/blob/7f192e7e395d6eaf8c8d1605ccdeb379d9bfe48d/app/javascript/components/widgets/selectors.js#L289.

In addition to this @aabdaab is correct that this will affect the map analysis. When the widget isn't rendered on the map, as the layer has an analysisConfig associated with it, on the fly analysis will be triggered to show a single analysis value instead. Is this wanted? We need to get very clear instructions from WRI about when to show this widget and its data. 
1. Is the country pre calc data good enough to show at any level? Or should we manually verify the country size before render
2. Is the pre calc data for AOIs good too? If so then we make sure we don't show this widget until the area is `status: 'saved'`
3. Is it just the showing of the layer next to the widget that is the problem? If so then we can disable the map layer in the dashboard and keep the widget, and not show the widget on the map by changing the `visible` prop of the widgets

edbrett:
@aabdaab @blayhem here you have an issue with analysis config on the layer level: https://bit.ly/2R6iOPA. This doesn't require any change in the code, so the PR is good. However you need to delete the `analysisConfig` from the layer `applicationConfig` in the API before release

app/javascript/components/widgets/forest-change/tree-loss-tsc/index.js
```diff
@@ -57,8 +57,8 @@ export default {
     },
     // loss tsc
     {
-      dataset: '89755b9f-df05-4e22-a9bc-05217c8eafc8',
-      layers: ['fd05bc2c-6ade-408c-862e-7318557dd4fc']
+      dataset: 'f23a1803-ba17-4fe3-a1ad-39abfbfb56c6',
```
aabdaab, line 60: hey, now all the IDs of the datasets and layers that are used somewhere in the codebase are moved to `layers-datasets.js` and `layers.js` files, could you move it to consts there? 🙏

edbrett: ah great. will do

edbrett: Hmm this means it wont work on staging as they will have different IDs.

aabdaab: we don't have dedicated datasets and layers for staging and production, these are the layers that we have: https://docs.google.com/spreadsheets/d/1V0tyqGoqOIfzMwx8f0Ix0-KExM1I63wCgUgbPoCW708/edit#gid=148980076

edbrett: Ok good to know. Updated.

PR #3937, opened 2020-04-03
Fix/add location path to located widget list

edbrett:
Loss in location widget paths weren't defined. Adds them back


edbrett:
> Add the same guard as for the `label` key `path: (region && region.path) || '',`, otherwise for some regions it crashes the page

Muy hecho 💆

PR #3940, opened 2020-04-03
Feature/limit aoi size

edbrett:
## Overview

Adds a new check in the analysis map view that validates the size of a geostore is less than 1 billion hectares. If true the `save in my gfw` button which opens the `save area modal` is disabled and displays a tooltip explaining why. Only applies when looking at a custom area.

## Demo

Screenshot 2020-04-03 at 15 06 53

## Testing

- draw a massive area and try and save it
- analyze a country, wdpa, use and make sure you can still save those

### DIFF
--- a/app/javascript/components/analysis/selectors.js
+++ b/app/javascript/components/analysis/selectors.js
@@ -28,6 +28,8 @@ const selectEmbed = state =>
   state.location.pathname.includes('/embed');
 const selectError = state => state.analysis && state.analysis.error;
 const selectDatasets = state => state.datasets && state.datasets.data;
+const selectGeostoreSize = state =>
+  state.geostore && state.geostore.data && state.geostore.data.areaHa;
 
 export const getLoading = createSelector(
   [
@@ -172,6 +174,17 @@ export const getLayerEndpoints = createSelector(
   }
 );
 
+export const checkGeostoreSize = createSelector(
+  [selectGeostoreSize, getDataLocation],
+  (areaHa, location) => {
+    if (!['aoi', 'geostore'].includes(location.type)) {
+      return false;
+    }
+
+    return areaHa > 100000000;
+  }
+);
+
 export const getAnalysisProps = createStructuredSelector({
   loading: getLoading,
   error: selectError,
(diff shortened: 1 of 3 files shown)


app/javascript/components/analysis/selectors.js
```diff
@@ -172,6 +174,17 @@ export const getLayerEndpoints = createSelector(
...
 );
 
+export const checkGeostoreSize = createSelector(
+  [selectGeostoreSize, getDataLocation],
+  (areaHa, location) => {
+    if (!['aoi', 'geostore'].includes(location.type)) {
+      return false;
+    }
+
+    return areaHa > 100000000;
```
aabdaab, line 184: can we reverse the condition? negation conditions tend to break one's brain 🤕 
```
if (isAOI || isGeostore) {
   return areaHa > 100000000;
}

return false;
```

app/javascript/components/analysis/selectors.js
```diff
@@ -28,6 +28,8 @@ const selectEmbed = state =>
   state.location.pathname.includes('/embed');
 const selectError = state => state.analysis && state.analysis.error;
 const selectDatasets = state => state.datasets && state.datasets.data;
+const selectGeostoreSize = state =>
+  state.geostore && state.geostore.data && state.geostore.data.areaHa;
```
aabdaab, line 32: not saying you should use it, just as a curiosity, there's a lodash `get` helper that lets you access deeply nested props (and if one of them at some point is `null` or `undefined` it'll return undefined): https://masteringjs.io/tutorials/lodash/get

app/javascript/components/analysis/selectors.js
edbrett: Totally correct. I only hope you can forgive me.

aabdaab:
Check my comments, otherwise looks good 👍 At some point would be top to do the same for widgets perhaps?

app/javascript/components/analysis/selectors.js
edbrett: I didn't know about that! Very cool. I think as we are doing the same across the site then not worth changing now, but will try to start using. The babel 7 updates for ?. chaining I guess do the same thing.

edbrett:
Updated. Going to merge.

PR #3944, opened 2020-04-06
Show newly created areas on the top of the list - map menu my-gfw section

aabdaab:
This PR changes the order of user areas on the map menu. It will move the newly created areas to the top of the list based on the area's `status` prop (newly created have status `pending`)

## Testing

- Go to the map and draw a custom shape
- Save the area, give it a `zzzz` name (`saved` areas are ordered by name)
- Note that this area will be listed as the first one on the list
- New areas will be ordered by `createdAt` date

reorderedAreas

### DIFF
--- a/app/javascript/components/map-menu/components/sections/my-gfw/selectors.js
+++ b/app/javascript/components/map-menu/components/sections/my-gfw/selectors.js
@@ -1,5 +1,6 @@
-import { createStructuredSelector } from 'reselect';
+import { createSelector, createStructuredSelector } from 'reselect';
 import isEmpty from 'lodash/isEmpty';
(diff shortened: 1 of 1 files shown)


aabdaab:
This is not a great solution after all. We're not taking into account the saved areas like countries, regions etc. (the ones that are already `saved`)

aabdaab:
I'm setting this as blocked as this needs feedback from a designer to agree on the best UI solution. We discussed with Luisa that this should be addressed in the next phase (part 1.1).

edbrett:
Ok!

edbrett:
Closing as we have found an alternative solution for now.

PR #3946, opened 2020-04-07
Change sublabel text in the multi select component

aabdaab:
Change the sublabel text in select component with `multiselect` option to a more accurate one. (For Mac users it's not the `control` button but the `command`)


aabdaab:
I know, such an improvement 😂

PR #3953, opened 2020-04-13
'Other' Input component, conditional label, required fields, other

blayhem:
## Overview

- Makes language required always in the save area form.
- Hides the label from the form Input if there isn't any (and adds a prop class to the Input)
- Radio input now uses Input instead of Field
- Styled radio input to add indentation


app/javascript/components/forms/components/input/component.jsx
```diff
@@ -61,7 +63,7 @@ class Input extends PureComponent {
               />
             ) : (
               <input
-                className="c-form-input"
+                className={`c-form-input ${className}`}
```
edbrett, line 66: lets use `classnames` here as with other components.

PR #3959, opened 2020-04-15
Added check for subsector in save AOI modal

blayhem:
## Overview

This fixes a bug where someone clicked "edit area" from the MyGFW page, the first page they saw is "my profile" page (see staging).

The reason behind this is that we didn't check for sub-sectors other than "Other: <something>". This caused the profile form to be considered "not filled" all the time if you selected another sub-sector.

To test:

1. Go to staging/my-gfw, try to update your profile with a sub-sector other than "Other", then try to edit an area.
2. Go to the branch deployment and try the same flow.
3. Also check the save AOI modal in the map.

### DIFF
--- a/app/javascript/components/modals/area-of-interest/component.jsx
+++ b/app/javascript/components/modals/area-of-interest/component.jsx
@@ -43,6 +43,7 @@ class AreaOfInterestModal extends PureComponent {
       !!email &&
       (!!fullName || !!lastName) &&
       !!sector &&
+      subsector &&
       (subsector.includes('Other')
         ? // if it's 'Other: <input>', we make sure that the input is not empty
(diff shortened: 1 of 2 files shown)


app/javascript/components/modals/area-of-interest/component.jsx
```diff
@@ -44,8 +44,10 @@ class AreaOfInterestModal extends PureComponent {
       (!!fullName || !!lastName) &&
       !!sector &&
       (subsector.includes('Other')
```
edbrett, line 46: If its undefined what will happen?

blayhem: good catch, thanks, it should be ok now

PR #3967, opened 2020-04-21
TCL 2019 Update

01painadam:
## Todo

- [x] Update Staging Tables
- [x] Update Loss widgets to allow 2019
- [ ] Remove Primary Forest from Loss Widget
- [ ] Add Primary Forest Loss Widget
- [ ] QA Loss widgets data:
    - [x] Adm0
    - [ ] Adm1
    - [ ] Adm2
- [ ] Update map loss layer (and decode)
- [ ] Update map biomass loss layer (and decode)

## Other updates

- Updates TCS Loss widget to automatically fetch the endYear from data

## Issues

- TCS Driver data is broken (all values are `"Unknown"`)


app/javascript/components/widgets/forest-change/tree-loss-tsc/index.js
```diff
@@ -95,21 +95,20 @@ export default {
...
 
-        const { startYear, range } = getYearsRange(data.loss);
+        const { startYear, endYear, range } = getYearsRange(data.loss);
 
         return {
           ...data,
           settings: {
             startYear,
-            endYear: 2018
+            endYear
```
edbrett, line 111: what happens if the data isnt zero filled? maybe not a problem

PR #3977, opened 2020-04-28
[FIX] Fixed /howto extLink in about page and TLD length in contact-us form.

blayhem:
## Overview

- Fixed an issue (probably related to the router?) in which the /howto link in the About page showed an error. Changed this link into an external link (opening a new tab) to gfw.org/howto.
- Changed the email validator to allow up to 10 character top level domains. Longest TLD every is way longer, but this should be enough for now...

### DIFF
--- a/app/javascript/components/forms/validations.js
+++ b/app/javascript/components/forms/validations.js
@@ -7,7 +7,7 @@ export const number = value =>
   (value && isNaN(Number(value)) ? 'Must be a number' : undefined);
 
 export const email = value =>
-  (value && !/^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}$/i.test(value)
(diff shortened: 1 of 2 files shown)


app/javascript/components/forms/validations.js
```diff
@@ -7,7 +7,7 @@ export const number = value =>
   (value && isNaN(Number(value)) ? 'Must be a number' : undefined);
 
 export const email = value =>
-  (value && !/^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}$/i.test(value)
+  (value && !/^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,10}$/i.test(value)
```
edbrett, line 10: What is the reason for the regex change?

blayhem: Allowing top level domains up to 10 chars long (.earth and longer)

edbrett: Muy bien

PR #3983, opened 2020-05-06
Fix/aoi cards fetch

blayhem:
## Overview

This fixes an issue when fetching fires data for the AOI cards when one of the two requests failed, or the area was pending.


app/javascript/components/aoi-card/component.jsx
```diff
@@ -128,6 +130,7 @@ class AoICard extends PureComponent {
 
     const isSubscribed = deforestationAlerts || fireAlerts || monthlySummary;
     const subscribedToAll = deforestationAlerts && fireAlerts && monthlySummary;
+    const isPending = status === 'pending';
```
edbrett, line 133: We should also stop the fetch from being triggered on mount.

blayhem: Good idea, done 👍

PR #3984, opened 2020-05-06
Added a select input for multiple select form component

blayhem:
## Overview

[Reference here](https://basecamp.com/3063126/projects/10727890/todos/415117549). This PR adds a write-in box for the multi-select option when 'Other' is selected.

image


blayhem:
> What happens in the post request?

Fixed, thanks!

edbrett:
This isnt saving anything in the API therefore updating the state of the user after saving. Please make sure it is saving the data.

PR #3986, opened 2020-05-12
Feat/fires all years test

01painadam:
An implementation of multiple years in the Cumulative widget, introducing an 'All' option in the compare years selector. Introduces a hideous tooltip on MODIS.

Screen Shot 2020-05-14 at 15 50 38

### DIFF
--- a/app/javascript/components/widgets/forest-change/fires-alerts-cumulative/selectors.js
+++ b/app/javascript/components/widgets/forest-change/fires-alerts-cumulative/selectors.js
@@ -404,7 +404,8 @@ export const parseSentence = createSelector(
     const maxWeek = maxBy(raw_data, 'count');
     const maxTotal = maxWeek.count;
     const maxYear = maxWeek.year;
-    const total = maxBy(slicedData, 'count').count || 0;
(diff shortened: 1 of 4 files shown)


app/javascript/components/widgets/forest-change/fires-alerts-cumulative/selectors.js
```diff
@@ -177,12 +184,12 @@ export const getLegend = createSelector(
         label: `${moment(end.date).format('YYYY')}`,
         color: colors.main
       },
-      ...(compareYear && {
-        compare: {
-          label: `${compareYear}`,
-          color: '#49b5e3'
-        }
-      }),
+      // ...(compareYear && {
```
edbrett, line 187: comments?

01painadam: Yeah - it needs refactoring to iterate through the array of compare years so that it works whether a single year, or all years are selected.

01painadam: Removed

PR #4006, opened 2020-06-02
Alphabetically sorted aois.

blayhem:
## Overview

Fixes [this issue](https://basecamp.com/3063126/projects/10728552/todos/418214819). Areas of interest are now sorted alphabetically in myGFW page and myGFW section in the map.


blayhem:
> Looks ok. Can you check the map sort order. The map panel should be sorted by createdDate but it seems to have broken.

I think it's working, but I also think they requested it to be alphabetically sorted. I just restored the "sort by date" bit.

edbrett:
> > Looks ok. Can you check the map sort order. The map panel should be sorted by createdDate but it seems to have broken.
> 
> I think it's working, but I also think they requested it to be alphabetically sorted. I just restored the "sort by date" bit.

Yes but in the map case the specifically want it sorted by created date

PR #4015, opened 2020-06-09
Add uppercase key to land categories

01painadam:
## Overview

Adds a general case check for all indicators.

If `upperCase =True` in the forestType or landCategory, then dynamic sentence preserves the original string (no lower casing).

Also, adds `dataType='glad'` in order to remove gain discalimer from the widget.

### DIFF
--- a/app/javascript/components/widgets/forest-change/glad-alerts/index.js
+++ b/app/javascript/components/widgets/forest-change/glad-alerts/index.js
@@ -34,7 +34,7 @@ export default {
   colors: 'loss',
   chartType: 'composedChart',
   source: 'gadm',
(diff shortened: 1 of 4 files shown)


app/javascript/components/widgets/forest-change/glad-alerts/index.js
```diff
@@ -34,7 +34,7 @@ export default {
   colors: 'loss',
   chartType: 'composedChart',
   source: 'gadm',
-  dataType: 'loss',
+  dataType: 'glad',
```
edbrett, line 37: dataType is only used for the footer statements so this is ok.

PR #4019, opened 2020-06-10
Fix / unitless pct lollipops

blayhem:
## Overview

Fixing rounding on small numbers (percentages) for the lollipop widget.

image

### DIFF
--- a/app/javascript/components/charts/lollipop-chart/component.jsx
+++ b/app/javascript/components/charts/lollipop-chart/component.jsx
@@ -3,8 +3,8 @@ import MediaQuery from 'react-responsive';
 import Link from 'redux-first-router-link';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
+import { format } from 'd3-format';
 
-import { formatNumber } from 'utils/format';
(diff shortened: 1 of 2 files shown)


app/javascript/components/charts/lollipop-chart/component.jsx
```diff
@@ -38,6 +38,16 @@ class LollipopChart extends PureComponent {
           : selectedUnitConfig.value;
     }
 
+    // TODO: pass as prop ?
+    const customFormat = ({ num, roundTo }) => {
+      if (num === 0) return '0';
+      let number = format(`.${roundTo || 2}f`)(num);
+      if (number.charAt(number.length - 1) === '0') {
```
edbrett, line 45: Im not sure I understand what is happening here. Can you explain and give some examples?

blayhem: Sure! It just removes trailing zeros i.e. 0.70 to 0.7.

edbrett: Ah I see it now. Thanks.

PR #4021, opened 2020-06-11
Feat/climate widgets to tables

01painadam:
## Overview

A number of climate widgets currently do not use the data tables, this process swaps them over from carto fetches.

No enhancements.

### Todo

- [x] Check downloads work as expected
- [x] Links on Rank widget
- [x] remove old carto fetch code (?)


edbrett:
I have pushed a commit to address the above comment

PR #4025, opened 2020-06-15
Adds primary forest to map for primary loss widget

01painadam:
## Overview

Adds primary forest to dashboard map for Primary Loss widget.

### DIFF
--- a/app/javascript/components/widgets/forest-change/tree-loss-pct/index.js
+++ b/app/javascript/components/widgets/forest-change/tree-loss-pct/index.js
@@ -108,8 +108,7 @@ export default {
   },
   settings: {
     threshold: 30,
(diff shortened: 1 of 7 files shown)


app/javascript/components/widgets/forest-change/tree-loss-pct/index.js
```diff
@@ -75,6 +77,11 @@ export default {
     {
       dataset: FOREST_LOSS_DATASET,
       layers: [FOREST_LOSS]
+    },
+    // primary forest
+    {
```
edbrett, line 82: You dont want to add this block. By having the forstType: 'primary_forest' you automatically add the layer. Having this adds the layer twice.

PR #4027, opened 2020-06-15
Feat/exlc emissions

01painadam:
## Overview

The default Loss widget fetch includes emissions keys (`whrc_aboveground_biomass_loss__Mg`, and `whrc_aboveground_co2_emissions__Mg`) however it is unused in the case of the Primary Loss widget.

Within `params` we can now pass an `excludeEmissions` key to omit the emissions keys from the fetch.


edbrett:
Im going to close this but leave the branch as it isn't relevant right now and adds some uncertainty with fires coming up.

PR #4050, opened 2020-07-23
Account for period that overlaps years

01painadam:
## Overview

Updates ranked fires widget's sql and selectors to account for user selected time periods that wrap around a calendar year (e.g. Nov 2019 to Feb 2020)

### DIFF
--- a/components/widgets/fires/fires-ranked/selectors.js
+++ b/components/widgets/fires/fires-ranked/selectors.js
@@ -38,23 +38,40 @@ export const getYears = createSelector([getLatestDate], latest => {
   return years;
 });
 
+export const getFilterWeeks = createSelector([getLatestDate, getOptionsSelected], (latest, options) => {
+  const period = options.weeks && options.weeks.value || 4;
+  const endWeek = moment(latest).isoWeek();
+  const startWeek = moment(latest).subtract(period, 'weeks').isoWeek();
+  return {startWeek, endWeek};
+});
+
 export const getStatsByAdmin = createSelector(
-  [getData, getYears, getAdm1, getLocation],
-  (data, years, adm1, location) => {
+  [getData, getYears, getAdm1, getLocation, getFilterWeeks],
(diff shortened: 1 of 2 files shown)


components/widgets/fires/fires-ranked/selectors.js
```diff
@@ -38,23 +38,40 @@ export const getYears = createSelector([getLatestDate], latest => {
...
     const filteredAlertsByAdmin = Object.entries(alertsByAdm).map(
       ([adm, adminAlerts]) => {
-        const countsArray = years.map(year => {
-          const filteredYear = adminAlerts.filter(el => el.year === year);
-          return filteredYear.length > 0 ? sumBy(filteredYear, 'count') : 0;
-        });
-
+        let countsArray = [];
+        if (startWeek < endWeek) {
+          countsArray = years.map(year => {
```
edbrett, line 62: is it possible for any of these to be undefined?

components/widgets/fires/fires-ranked/selectors.js
```diff
@@ -38,23 +38,40 @@ export const getYears = createSelector([getLatestDate], latest => {
...
+            const filteredYear = adminAlerts.filter(el => el.year === year);
```
edbrett, line 63: everything below here is repeated so you could place the if else inside the `.filter`

components/widgets/fires/fires-ranked/selectors.js
01painadam: startWeek and endWeek have a fallback.

components/widgets/fires/fires-ranked/selectors.js
01painadam: I'm avoiding having to do logic on every iteration of the map. If the period between start week and end week doesn't cross a year then theres no need to filter. In the case of country or global dashboard there will be thousands of elements in `adminAlerts` so I want to avoid filtering if it's not necessary so we have decent performance.

PR #4063, opened 2020-08-05
fires layers needs some filters to be parsed before render

tomasmoose:
## Overview

Using the Mapbox `"in"` expression requires us to be able to pass an array of  strings (or numbers, potentially) to the render. Currently this is unsupported on our end prior to passing the config to the Layer Manager library.

This feature not spreads arrays passed by `params_config` into the render by default.

e.g.

```
"params_config": [
{
"default": [
"h",
"n",
"l"
],
"key": "confidence_level",
"required": true
}, ...
```
passed into render with:
```
[
"in",
"confidence__cat",
"{confidence_level}"
]
```

becomes:

```
[
"in",
"confidence__cat",
"h",
"n",
"l"
]
```

## Notes

Fires layer: https://api.resourcewatch.org/v1/layer/2f1f6c86-bef8-4f58-b146-ee51e228a7e0?includes=vocabulary,metadata

### DIFF
--- a/components/map/selectors.js
+++ b/components/map/selectors.js
@@ -12,24 +12,26 @@ import { getActiveArea } from 'providers/areas-provider/selectors';
 import basemaps from './basemaps';
 
 // map state
-const getMapSettings = state => state.map?.settings || {};
-const selectMapLoading = state => state.map && state.map.loading;
-const selectGeostoreLoading = state => state.geostore && state.geostore.loading;
-const selectLatestLoading = state => state.latest && state.latest.loading;
-const selectDatasetsLoading = state => state.datasets && state.datasets.loading;
-const selectRecentImageryLoading = state =>
+const getMapSettings = (state) => state.map?.settings || {};
+const selectMapLoading = (state) => state.map && state.map.loading;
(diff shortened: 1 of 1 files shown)


components/map/selectors.js
```diff
@@ -426,43 +440,82 @@ export const getActiveLayers = createSelector(
...
-    return layers.map(l => {
+    return layers.map((l) => {
       const { decodeFunction, decodeParams } = l;
       const { startDate, endDate } = decodeParams || {};
+      const { layerConfig } = l;
+      const { params_config = [] } = layerConfig;
+
+      // Some filters requires values to be spread
+      // this modifies the filters that has those requirements to return all values
+      const requiresSpread = params_config.filter((config) =>
```
01painadam, line 453: Nice solution - I guess the question is should we flag this in the `params_config` on the layer with `is_spread_key: true` or something?

In the mapbox vector styles regime in theory it will always  be  a string or an array that we pass in to the render... so maybe this solution is enough.

Opinions, @edbrett ?

edbrett:
closing this as we no longer require a solution for this.

PR #4065, opened 2020-08-07
Feature/reduce url params

edbrett:
## Overview

The flow for serialising the store and mapping the keys to query params for the url was capturing the entire possible state. This meant that urls were becoming extremely long. This checks the query params against the initial state to reduce this.

## Testing

- Go to the map, move around, check url, reload and make sure you are returned to the same state
- Repeat for dashboards
- Repeat for my gfw.


edbrett:
https://gfw-staging-pr-4065.herokuapp.com/

PR #4073, opened 2020-08-25
fix: disable download for glad

tomasmoose:
## Overview

This pr disables download button for glad, and adds a disclamer to widget/layer

Notion: https://www.notion.so/9dd860f5dfd64475b92f4a3f79f0ff37?v=2098a476c4fa407981718cb4eafad507&p=4831737872e64d7793cebb1fea1f354a

### DIFF
--- a/components/widget/components/widget-header/component.jsx
+++ b/components/widget/components/widget-header/component.jsx
@@ -36,6 +36,7 @@ class WidgetHeader extends PureComponent {
 
   render() {
     const {
+      widget,
       title,
       loading,
       active,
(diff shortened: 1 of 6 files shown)


elpamart (changes requested):
The caveat text on the map should be the shorter version and the download option for the Deforestation alerts in XXX (dashboard), the Location of deforestation alerts in XXX (dashboard) and the Deforestation alerts in XXX (map) has to be grayed out with a popup saying 'Temporarily unavailable' instead of being removed.

elpamart (commented):
Looks good to me!

PR #4082, opened 2020-09-07
Adds static decode/encode remapping

01painadam:
## Overview

Adds a generic shader static remapping function to be used with any rgb tile layer.

Used to remove blurred edges from rgb tile cache layers at the (small) expense of performance.


edbrett (approved):
A useful addition to the decode gang 🤝

PR #4083, opened 2020-09-08
Feature/map search

tomasmoose:
## Overview

This pr adds new design update for the search panel. 

1. there are a few style changes for previous content, but basically kept the colors the same as defined colors where not in our config. 
2. Modified styles to make sure only search results are overflowed, otherwise the entire menu will move when results overflow the container
3. As previous statement, i decided to make sure the search box is always visible on desktop when scrolling, so you dont loose context of what you searched for. 
4. Mobile layout is handled mostly by flex logic, as we will only show eather menu/result container and let flex grow deal with the layout. For this change the logic became a bit more complex in the search component; so did some splitting of logic there to make it easier to work with in the future. 
5. Created a `vertical-menu` component that we can re-use when we need it, as of now its quite simple and we can extend if we need more logic there. 

## Demo

Screenshot 2020-09-08 at 14 42 47

## Notes

We still use "one" component for datasets/locations, as the behaviour is still similar. and when switching between the location/dataset menu it behaves like a "filter" and you dont loose the previous context. 

...

## Testing

- Make sure search works as it was before, as nothing should have been changed there, but all old logic should still work. 
- Make sure navigation works as expected.
- Make sure styles are correct (as i didnt use the colors in (invision). 
- Tested in webkit,ff edge not sure if we need to check ie 11 ? 
- Make sure mobile/desktop layouts behave as expected both in reload/resize browser window.


components/map-menu/components/sections/search/component.jsx
```diff
@@ -11,45 +11,64 @@ import UTMCoords from './components/utm-coords';
...
+        onClick={value => setMenuSettings({ searchType: value })}
+      />
+    )
+  }
+
   render() {
-    const { searchType, setMenuSettings, isDesktop } = this.props;
+    const { searchType, isDesktop } = this.props;
+
+    // Mobile view displays eater menu or container depending on current action
```
edbrett, line 48: spelling mistake? `eater` -> `either` ?

components/ui/vertical-menu/component.jsx
```diff
@@ -0,0 +1,53 @@
...
+              <Icon icon={arrowIcon} className="arrow-icon" />
+            </Button>
+          </li>
+        ))}
+      </ul>
+    )
+  }
+}
+
+VerticalMenu.propTypes = {
```
edbrett, line 43: As we are using PureComponent this could be called as a `static` on the class to be more efficient.

edbrett (commented):
Nice job! Good simplification of the search panel with class methods. Much easier to read. 

Only feedback is for the Vertical menu component, the `button` does not fill to whole area of the menu item due to the padding being on the parent wrapper. Could we move this inside the button to improve the interaction? 

Screenshot 2020-09-10 at 09 38 08

edbrett:
There is a whole other issue of how to access this menu on mobile but we can address that is another PR.

components/map-menu/components/sections/search/component.jsx
tomasmoose: > There is a whole other issue of how to access this menu on mobile but we can address that is another PR.

yeah saw that, though it was a feature :D guess we can add it back to the menu in another pr!

components/ui/vertical-menu/component.jsx
tomasmoose: good catch, i moved this in another component aswell `MapMenuSearch`

tomasmoose:
fixes amended ! @edbrett

tomasmoose:
added an issue to notion related to search on mobile: [here](https://www.notion.so/Bugs-47e054fc0652498d9fb555ca2070c238?p=a2618b6af8824de7842d5060430873cb)

components/ui/vertical-menu/component.jsx
edbrett: Thanks!

edbrett:
This is looking great! Im not going to merge yet as we are missing the results from geocoding service which we are still waiting for confirmation on. I will tag as on hold or similar.

tomasmoose (commented):
Great improvements @edbrett ! im approving!

PR #4084, opened 2020-09-10
Filter 2001 fromo totals

01painadam:
## Overview

Totals in the selectors now filter out the 2001 loss year.


edbrett (approved):
Good catch.

PR #4088, opened 2020-09-15
Feature/sentence tests

tomasmoose:
## Overview

This pr adds basic sentense verification e2e tests and logic to make it easier to add more on top of current ones 

### Notes

We are now using fixtures for some specified requests, this is to avoid latency when especially running on a CI environment. Inside init-spec-file we now have logic that if you run `cy:record` will record requests to a fixtures file. This will make sure we don't run into timeout issues and will speed up our e2e tests substantially. As these requests will on a ci environment be mocked. 

I also increased timeouts for some generic settings as GFW requires a substantially higher timeout than a "normal" application. 

Initial documentation can be found [here](https://www.notion.so/Front-end-testing-553214d114754d93a6bb3339b67857f6)

### To run tests

- Run `yarn cypress install`
- Run `yarn dev`
- in a separate command line once the application is running locally: `yarn cy:open`

## To be done 

- [x] Code cleanup 
- [x] Documentation


edbrett (commented):
I have no issues with these, it all seems good to me

edbrett (approved):
Fantastic work 🎖️ 🏅

PR #4091, opened 2020-09-23
Feature/add babel polyfill back for IE 11

edbrett:
## Demo

## Notes

## Testing

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,10 @@
 {
-  "presets": ["next/babel"],
-  "plugins": ["istanbul"]
+  "presets": [
+    "next/babel"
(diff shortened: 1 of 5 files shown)


tomasmoose:
At the moment packages that needs to be transpiled are. So you end up seeing the map on IE. Problem is we have some other shader/performance issues and probably lots of unknown issues that will not be a quick fix. so this pr is on hold for now.

Another note;

deck.gl will deprecate IE11 from version 8.

Already discussed on slack with ed;

tomasmoose:
I will close this, but we can keep the branch for a while while we migrate things

PR #4093, opened 2020-09-24
Feature/static map and dashboard pages

edbrett:
## Overview

Updates the map and dashboard routes to use next v9's `getStaticPaths` and `getStaticProps` when building the app. With this comes a few positive side effects:
1. we are now using the new nextjs [catch all](https://nextjs.org/docs/routing/dynamic-routes#catch-all-routes) which allows for a path of `/map/` to exist without needing a rewrite on the server to `/map/global/`. This solves the issues we were having with query params not updating and is more inline with what we are trying to achieve.
2. we can pre-build the country pages so that they are served statically, making them lighter and faster
3. we can now use the [next-sitemap](https://github.com/iamvishnusankar/next-sitemap) library to parse our statically built pages and dynamically output a sitemap to our public folder. This has some cool features:
- we can trigger out sitemap with the npm `postbuild` script allowing us to rebuild the sitemap whenever we deploy.
- the sitemap will always be up to date with the current page routes of the app
- a `robots.txt` files is output in parallel with the sitemap referencing any (multiple) sitemaps generated and with a size limit. This helps us to follow SEO best practices. 

## Notes

The `next-sitemap` library has the following config. Note the excluded paths and transform functions for handling exceptions to the rule.

```
module.exports = {
  siteUrl: 'https://www.globalforestwatch.org',
  generateRobotsTxt: true,
  sitemapSize: 5000,
  changefreq: 'weekly',
  exclude: [
    '/404',
    '/search',
    '/thank-you',
    '/browser-support',
    '/terms',
    '/my-gfw',
    '/privacy-policy',
    '/subscribe',
  ],
  transform: (config, url) => {
    if (url.includes('/embed')) {
      return null;
    }

    if (url.includes('/map') || url.includes('/dashboards')) {
      return {
        loc: url,
        changefreq: config.changefreq,
        priority: '0.7',
        lastmod: config.autoLastmod ? new Date().toISOString() : undefined,
      };
    }

    return {
      loc: url,
      changefreq: config.changefreq,
      priority: '1.0',
      lastmod: config.autoLastmod ? new Date().toISOString() : undefined,
    };
  },
};
```

## Testing

- check out the `/sitemap.xml`
- visit the global map `/map` and make sure query params are updating
- visit a country map page through click interaction
- visit the dashboards and change countries
- visit a widget embed page


components/widgets/selectors.js
```diff
@@ -78,7 +78,8 @@ export const selectEmbed = (state, { embed }) => embed;
 export const selectSimple = (state, { simple }) => simple;
 export const selectAnalysis = (state, { analysis }) => analysis;
 export const selectCategory = (state) =>
-  state.location && state.location.query && state.location.query.category;
+  (state.location && state.location.query && state.location.query.category) ||
```
tomasmoose, line 81: "not related", Do we have optional chaining supported in our build flow? would prefer this in the future to clean up these conditions! 

```
export const selectCategory = (state) => state?.location?.query?.category || 'summary';
```

edbrett: Yes we do! It default in babel 7 now!

tomasmoose: awesome! no need to update now, we can do it as we go along

tomasmoose (approved):
1. sitemap valid
2. Interactions and URL updates accordingly 
3. Embed works

Great work!

edbrett:
Thanks! Then off it goes, into the wild 🥏 🌴

PR #4095, opened 2020-10-01
Feature/basic mapbuilder page

edbrett:
## Overview

Adds the MapBuilder page from the developers portal to the flagship site as developers tools is being decommissioned. 

## Notes

The page uses the flagships components so some visuals are updated and more inline with the UI kit. 

## Testing

- go to `/mapbuilder` and review the content of the page


layouts/mapbuilder/config.js
```diff
@@ -0,0 +1,75 @@
...
+      text:
+        'Easily embed your MapBuilder into your own website, share the link with colleagues and peers or use integrated social media tools for sharing.',
+    },
+  ],
+  maps: [
+    {
+      title: 'Cameroon Forest Atlas',
+      summary: 'View an interactive map with land-use data for Cameroon.',
+      image: cameroonForestAtlasImg,
+      extLink: 'http://cmr.forest-atlas.org',
```
tomasmoose, line 59: as these links do not force a secure link we should make sure we redirect to a secure link from our side `https://cmr.forest-atlas.org`

layouts/mapbuilder/component.jsx
```diff
@@ -0,0 +1,153 @@
...
+          <div className="row">
+            <div className="column small-10 small-offset-1 medium-3 medium-offset-3">
+              <a href="mailto:gfw@wri.org">
+                <Button className="column-btn" theme="theme-button-light">
+                  submit your map
+                </Button>
+              </a>
+            </div>
+            <div className="column small-10 small-offset-1 medium-3 medium-offset-0">
+              <Link href="https://www.globalforestwatch.org/help/mapbuilder/">
```
tomasmoose, line 97: View all redirects to help center, if this is the case I would recommend some other copy so we are not misleading, even just reusing the above sections "get started" is fine

tomasmoose (approved):
Looking good! just two comments but nothing big! great work!

layouts/mapbuilder/config.js
edbrett: Fixed!

edbrett:
Thanks for the review. I have now integrated this with the CMS so we dont have to worry about content in the future.

PR #4097, opened 2020-10-09
Feat/refactor sentence tests

01painadam:
## Overview

Collects similar tests by spec under the assumption that all tests in that config use the same spec function.

Makes tests easier to describe within the tests key.


edbrett (commented):
these look reasonable also.

PR #4100, opened 2020-10-14
Check for type when inserting default value

01painadam:
## Overview

When default values are passed into the sql constructor, check type and add encapsulates with `'` as appropriate.

NOTE: this may need extending to all values.


services/analysis-cached.js
```diff
@@ -153,9 +154,9 @@ export const getWHEREQuery = params => {
         !tableKey.includes('is__') &&
         polynameMeta.default &&
         polynameMeta.categories
-    ? ` AND ${tableKey} ${polynameMeta.comparison || '='} '${
-      polynameMeta.default
-    }'`
+    ? ` AND ${tableKey} ${polynameMeta.comparison || '='} ${
+      polynameMeta?.dataType === 'keyword' ? `'${polynameMeta.default}'` : `${polynameMeta.default}`
```
edbrett, line 158: Nice. Do we have a fallback if .deafult isnt defined?

edbrett: Ping

edbrett (approved):

PR #4104, opened 2020-10-14
Feature/layers by slug

edbrett:
## Overview

Migrates the apps internal use of layer and dataset ids from the API to fixed slugs. 

**The problem**
When a layer or dataset id changes in the API the new entity with a new id will show up in the datasets menu and be able to be added to the map, however, as the reference to that entity has changed certain parts of the application will fail to work.
- if the id is in the url state then it will still pass down to the store on first load, but the datasets selectors will not find an entity and therefore the layer will not be active on the map or any other UI.
- when adding a widget to the map (dashboards or analysis), the reference for which layer to target and pass parameters to will no longer be valid and the selectors will fail
- any UI using layer id as a `value` reference for dispatching a change in state in the store, such as `selectorConfig`'s for the legend will follow the same failed results.

**The solution**
Within the applicationConfig of each layer in the API we have given a unique `slug` and `dataset_slug`. This slug is unique to the conceptual entity and not unique across all entities in the API. This gives us an extra layer of abstraction in the app where we can reference a layer or dataset throughout the app and not have to change this over multiple versions of that entity. We may also develop layers in tandem from an API level without worrying about the issues from above causing side effect when we eventually deploy and swap to the latest version.

To address this we need to update a few things:

- [x] update datasets provider to swap all references for ids to slugs in the store
- [x] update our constants files that reference layers and datasets to slugs
- [x] create a map of legacy ids and slugs so we can remap the urls from old links to point to the correct entities

## Notes

You may assign this slug to multiple layers or datasets but the application will only consider the first it finds. 

## Testing

- Load the map and add/remove layers, change their settings, all features should work.
- Load the dashboards and add a widget to the map. Change the widget settings, and check the map updates its layers
- Go to flagship and change some layers, copy the path and query to the staging branch for this PR, and check you get the same view.


edbrett:
Now updated and complete

tomasmoose:
hi @edbrett there seems to be no layers or datasets available in this environment, is this expected?

edbrett:
> hi @edbrett there seems to be no layers or datasets available in this environment, is this expected?

Nope 😓 I will check

tomasmoose (approved):
Code looks great; 

Tested: 

1. deleting/modifying etc works as expected
2. dashboards load on the map (check)
3. Flagship URL copying etc works great 

Great work!

01painadam (approved):
Everything functions as before - looks good to me 👍

tomasmoose:
@edbrett this is ready if you feel conformable merging lets do it!

PR #4110, opened 2020-10-26
Feature/split page wrapper components

edbrett:
## Overview

A hefty PR that encapsulates a wide range of small tasks that needed completing together to organise the whole app after the nextjs migration. #thoughts and #prayers 🙏 .

1. Add cope split page wrappers for different page types
Up until know, we have been using a single page wrapper component to handle the default styles, loaders and global actions of every page, such as changing language, tracking analytics etc. This has been split into appropriately scoped pages for each types: `page`, `fullscreen`, `static` and `embed`. There is also an updated `<Head />` component for handling CMS SEO tags and page level metadata. These can be found at `/layouts/wrappers`.

2. Simplify _app
The `_app` file for customising the default next wrapper contained two components, `<LocationProvider />` and `<MyGfwProvider />`, which serve no purpose outside of the pages that consume their data. These have been moved to their respective layouts for better code splitting. 

3. Pass static params correctly to `grants and fellowships` and `topics`
Both of these pages have been simplified significantly to abstract their data dependancies for projects to a service and `getStaticProps`. As a result the majority of the previous code is no longer required and the no longer have a redux dependancy. 

4. Remove redux from search page
The search page was over complicated and contained a small bug with setting of the `<input />` from the query change in the router. This has been refactored to be much simpler, relying only on query params to handle state changes and fetching.

5. Renames many folders and files to tidy redundant code
- The `./docs` containing the docasaurus app has been deleted as it isn't being used.
- `./data` and been renamed to `/docs` as it is much more sensible.
- `./static` has been renamed to `./data` as it is also even sensibler. 
- `./analytics` has been moved to utils.

6. Cookies and analytics
The cookies interactions have been improved to prevent the loading of tracking scripts until after acceptance. 

7. Lang helpers were moved to a scope utils file.

8. All instances on <Meda /> were migrated to using `gfw-components` as there will be breaking changes in the upcoming header update.

9. Many modals were simplified to remove redux and local state due to their simplicity.

In completing all of the above a variety of bugs were rehoused too:
- Topics cards links that contained selectors were fixed to show the correct defaults.
- Webhook url testing in the save area of interest modal now only allowed checking of a valid URL.
- Styles of the video modal improved to allow clear access to close button.
- Styles of all modals and static pages improved.

## Testing

- Visit each page and check that they render, the head has the correct title.
- Visit map, dashboards, and my-gfw logged in and access profile and areas modal.
- Perform a search from the header component and check the page input has the correct value. 
- Visit the site from incognito, check the network for `hotjar`, `facebook`, `twitter`, `crazyegg` and `google analytics`. It should not be present. Accept cookies and watch network for updated scripts.
- Change language and check that text updates.
- Open the area of interest modal and test a webhook url.
- Visit the climate topics pages, go to related tools, on the second card click the link to Brazil for the default dropdown choice.


tomasmoose (approved):
Great work!

PR #4116, opened 2020-10-28
Feature/simplify store

edbrett:
## Overview

Another clean up PR to start the process of simplifying our use of redux to supported libraries and tooling. The motivation for this is not only to reduce code, but also begin removing our dep on the [redux-actions](https://github.com/redux-utilities/redux-actions) lib which is no longer maintained.

The store has been reduced 😅 to a simple memoized create function, `makeStore`, which implements the [redux toolkit](https://redux-toolkit.js.org/) default store config for dev tools and middlewares. Previously we were handling to rehydration of the store from server side props which we never used, so the majority of the code was redundant. This brings some benefits:
- correctly configured dev tools and thunk out of the box
- less code
- `immutableStateInvariant` middleware for better reducer validation.

The adding of this additional middleware flagged errors in 2 actions:
1. `setMapInteractions` was passing `_vectorTileFeature` from mapbox to the store, a reference that is mutated by mapbox in view transitions and also of no use to use.
2. `getDatasets` default parsing of layer configs was generating an unfriendly rgba trackStyle color. This is now converted to a hex before returning.

## Testing

- Load the map and navigate between some locations. You should receive no redux related errors.

### DIFF
--- a/components/map/component.jsx
+++ b/components/map/component.jsx
@@ -136,7 +136,7 @@ class MapComponent extends Component {
         category: 'Map analysis',
         action: 'User opens analysis popup infowindow',
         label: interaction.label,
-      })
+      });
 
       if (interaction.data.cluster) {
         const { data, layer, geometry } = interaction;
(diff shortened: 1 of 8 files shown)


tomasmoose (approved):
Great work!

PR #4125, opened 2020-11-03
Fix/climate widget queries

edbrett:
## Overview

Elastic search updates in the API still have some int/string updates that we need to tweak on the front end queries for climate widgets.
- NYC cumulative glads widget queries year type INT
- Carbon stocks widget cannot query ISO.

BONUS:
- fixes for embed widgets failing when in global view
- style fix for the SVG brush
- move widget hover interaction outside of settings for better performance


01painadam:
Both widgets working now, looks good.

PR #4128, opened 2020-11-04
Feature/planet API proxy

edbrett:
## Overview

Add api route for proxy to planet tiles to hide API key in requests.

## Testing

- Go to the map and turn on planet basemaps, make sure the tiles load.
- Check the network for planet responses and check for API key


ikas (approved):
A simple solution for what could have been a big problem, love it! 

Looks great to me 👏 🚀

PR #4134, opened 2020-11-24
feat/otf-analysis

tomasmoose:
## Overview

Introducing new OTF analysis service for un-cached widget data.

## Notes

This service utilizes [this](https://data-api.globalforestwatch.org/#operation/zonal_statistics_get_analysis_zonal__geostore_id__get) API in its core and documentation of the actual implementation can be found [here](https://www.notion.so/OTF-analysis-service-59d1539be54e43b799ae3a783c56bcf7). 

To be done: 

1. Implement remaining data structures inside `data/otf-data.js`
2. Implement new service in all widgets that need OTF data. currently `forest-loss` implemented.

## Testing

- Ensure data is correct for new implementation (validate API) 
- Ensure widgets render correctly when OTF is needed.


data/otf-data.js
```diff
@@ -0,0 +1,11 @@
+export default {
+  loss: {
+    sum: ['area__ha'],
+    groupBy: ['umd_tree_cover_loss__year'],
```
edbrett, line 4: As discussed shall we try to find a way to abstract these to a constants file and reference them in data/analysis-cached, data/forest-types and data/land-categories?

tomasmoose: added a new file "temporary" where we can define these layers, will update the migration document of cached analysis where we will combine these next week! AJ will take a look and make sure we use the correct layers then I will patch all widgets that need this

PR #4153, opened 2021-01-11
OTF analysis

tomasmoose:
Overview
Introducing a new OTF analysis service for un-cached widget data.

Notes
This service utilizes this API in its core and documentation of the actual implementation can be found here.


01painadam (approved):
Works as intended, analyses return results that make sense for geostores.

tomasmoose:
@01painadam New data tag logic for e2e tests will be added to this pr, our widget tests will break atm, but they will be migrated on Monday!

PR #4157, opened 2021-01-14
Decode latest date from stats response

01painadam:
## Overview

Adds RADD Alerts specific `latestDate` parsing for the `endDate` param as well as the raster decode function `RADDsCoverage` to parse the RADD Alerts geographic coverage layer.

### Details

The `stats` endpoint for the RADD alerts dataset returns statistics (`min`, `max`, `histogram`) for the _encoded_ alert dates. Example response:

```
"data": {
    "bands": [
        {
            "min": 21462,
            "max": 32199,
            "mean": 31221.31386,
            "histogram": {
                "bin_count": 8638,
                "min": 21440.947059,
                "max": 32220.052941,
                "value_count": [ 1166, 89, 0 , 1455, ...
```

Statistics are returned in the format `abbbb` with:

- `a` = _Confidence level_ where: 3 is 'high confidence', 2 is 'nominal', 1 is 'low'
- `bbbb` = number of days since 2014-12-31.

### Calculations

Since stats are returned in their encoded form, the `max` value is **always high confidence** (since 3XXXX > 2XXXX or 1XXXX), therefore we need to do further parsing to extract the latest alert date _regardless of confidence level_.

To do this we need to look at the `value_count` array inside the `histogram_key`, which is a binned frequency histogram of all alerts that occurred inside that encoded bin. Since this is filled sequentially, in order to get the latest _encoded_ date, we need to find the index of the last non-zero element and multiply that by the bin size.

`bin_size = histogram.max - histogram.min / histogram.bin_count`

The latest encoded date is then: `histogram.min + lastNonZeroIndex*bin_size`.

Finally, the latest encoded date _that is not high confidence_ is the same as the above with the condition that the encoded date is < 30,000.

We then convert the larger of the two _decoded_ `bbbb` numbers into a latests date.

## RADD Extent Layer

Simple raster coverage decode added to config:

Screen Shot 2021-01-14 at 15 29 50

### DIFF
--- a/providers/latest-provider/actions.js
+++ b/providers/latest-provider/actions.js
@@ -25,6 +25,58 @@ export const getLatest = createThunkAction(
               responses.reduce((obj, response, index) => {
                 const latestResponse = response.data.data || response.data;
                 let date = latestResponse.date || latestResponse.max_date;
+                const { bands } = latestResponse;
+                // if the response is from the stats endpoint, get bands key
+                if (bands && bands.length) {
+                  const { max, histogram } = bands[0];
+
+                  // high confidence alerts from top-level max
+                  // max = abbbb where a = confidence value, bbbb = days since 2014-12-31
+                  const encodedLatestDate = max.toString(10);
+                  const daysSinceString = encodedLatestDate.substring(
+                    1,
+                    encodedLatestDate.length
+                  );
+                  const daysSince = parseInt(daysSinceString, 10);
+
+                  // low confidence alerts from histogram
+                  // need to find the last non-zero bin where where index < 30,000 (high conf)
+                  const binSize =
+                    (histogram.max - histogram.min) / histogram.bin_count;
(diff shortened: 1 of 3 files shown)


tomasmoose (approved):

providers/latest-provider/actions.js
```diff
@@ -33,9 +33,11 @@ export const getLatest = createThunkAction(
                   // Can this service return "null" or similar and then we can handle that case here
                   const days = statsLatestDecoder(bands);
                   // convert to date
-                  date = moment('2014-12-31')
-                    .add(days, 'days')
-                    .format('YYYY-MM-DD');
+                  date = days
+                    ? moment('2014-12-31')
```
tomasmoose, line 37: could we add this date as a constant @01painadam ?

01painadam: yes, it can

PR #4161, opened 2021-01-28
Update title to flux

01painadam:
## Overview

Changes Climate tab sub-section title from `Carbon Emissions` to `Climate Flux`

Screen Shot 2021-01-28 at 11 10 47

## Notes

The Dataset Vocabularies remain `categoryTab: carbonEmissions` we will have to change this in the future, but for now let's keep this update simple.


01painadam:
Tests passed - merging since its such a low risk change

PR #4164, opened 2021-01-28
fix planet basemap date/image type serialisation

tomasmoose:
## Overview

Fix multiple issue with planet basemaps. 

1. Use dates from basemap response instead of parsing title
2. Filter basemaps based on selected image type

## Notes

BC: https://basecamp.com/3063126/projects/10728552/todos/433532321

## Testing

- Ensure dates + basemap corresponds to the correct image type
- Ensure basemaps render, or make sure this is not an issue on our side 
- Check interactivity and lets make sure we are updating the basemaps dynamically correct
- Make sure that when you land on planet basemap selected you get results 
- Make sure that when you toggle between image types that basemap renders correctly
- Ensure date ranges within the basemap options are correct and all of them works as expected (for both image types)

### DIFF
--- a/components/basemaps/index.js
+++ b/components/basemaps/index.js
@@ -19,6 +19,7 @@ class BasemapsContainer extends React.Component {
   static propTypes = {
     activeLabels: PropTypes.object,
     basemaps: PropTypes.object,
+    defaultPlanetBasemapsByCategory: PropTypes.object,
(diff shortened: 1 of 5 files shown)


01painadam (approved):
Dates are fixes and returning what I'd expect for NIR and Natural colour categories - seems the Planet API is returning some blank tiles in the case of Natural colour, though.

PR #4167, opened 2021-02-02
fix/force-ssl

tomasmoose:


ikas (approved):
LGTM, and Heroku build is working fine (which I believe would be the major pain point with changing the `npm start` script), so I think this is good to be merged :+1:

PR #4180, opened 2021-02-24
add ssr support for dashboards

tomasmoose:
## Overview

This pr adds ssr support for dashboards we will on the server fetch these resources: 

1. Anything country related, removing 4x calls on initial dashboard load
2. Initial data for geodescriber needed to parse the sentence removing 3x+ calls on initial load

New parsing added:

1. We now parse the sentence on the backend, so when initially loading the page we will have the sentence ready. This is true for all admin levels
2. Fetch necessary country info for pre-populating admin selectors.

Removed calls

Removed on mount load countries and geo described info for parsing the sentence

## Testing

- Test dashboards, click around it should work as expected when switching admins etc.
- Test initial load for all admin levels, global/adm1/adm2 to make sure we pre-populare the dashboard header correctly.


tomasmoose:
Added a new embed page that will allow us to test all widgets in isolation. also deployed a fix that solved the location context on the server, this allows us to generate the correct context before we serve the page. 

This is the URL structure for the new dashboard header embed: `embed/sentence/[location]/[adm0]/[adm1]/[adm2]`

It mimicks the dashboards so should be quite easy to add new contexts to our tests for this. This new embed it wont be used for any users, only internally by us, also included data used to generate the sentence, this will be useful for automation + manual testing for any type of admin. 

Screenshot 2021-02-25 at 11 04 16

01painadam (approved):
Country and AoI dashboards working for me.

PR #4184, opened 2021-03-03
Make overflow visible for trase widget

Bluesmile82:
## Overview

On the [Trase integration](https://trase.earth/profiles) we had a problem with two scrollbars on the widget. It could only be solved by changing these styles inside the GFW application.
image

## Testing

We will need this to be deployed to staging/production to test it 😕


Bluesmile82:
This is working fine with the generated url
image

tomasmoose:
will deploy this on Monday as deployments are blocked this week, added blocked tag

tomasmoose (approved):

PR #4194, opened 2021-03-15
Planet Basemaps Test

01painadam:
## Update

We need to implement this in a way that allows it to work when there are Visual basemaps (using the staging key) and when there are not (using the current production key).

The user should always see2 options:

1. Natural Colour (`proc=''` if visual basemaps are available, else `proc='rgb'` and analytic basemap if not)
2. False Colour (NIR) (always `proc='cir' and analytic basemap`)

## Overview

Adds new code to test new visual basemaps for Planet. Note that is should fix a previous issue where visual basemaps were not returning correctly.

New basemaps have the the following nomenclature:

- `planet_medres_visual_20YY-MM_20YY-MM_mosaic` for 6-monthly mosaics
- `planet_medres_visual_20YY-MM_mosaic` for monthly mosaics

Here 'visual' basemaps need to be sent with the query param `proc=''`, whereas analytic mosaics can be sent with `proc='rgb'` or `proc='cir'` for 'false colour'.

This means that we have 3 options for each time range:

- Visual with `proc=''` (True colour)
- Analytic with `proc='rgb'` (False colour)
- Analytic with `proc='cir'` (False colour NIR)

## Note

Requires a new Planet API key in `.env`: `NEXT_PUBLIC_STAGING_PLANET_API_KEY` which is used when `NEXT_PUBLIC_FEATURE_ENV === 'staging'`.

### DIFF
--- a/components/basemaps/index.js
+++ b/components/basemaps/index.js
@@ -31,10 +31,12 @@ class BasemapsContainer extends React.Component {
 
   handlePlanetName = (name, color) => {
     const { defaultPlanetBasemapsByCategory } = this.props;
-    const { visual, cir } = defaultPlanetBasemapsByCategory;
+    const { visual, cir, rgb } = defaultPlanetBasemapsByCategory;
     if (!name) {
       // User selects image category
-      return color === 'cir' ? cir : visual;
+      if (color === 'rgb') return rgb;
+      if (color === 'cir') return cir;
+      return visual;
     }
     return name;
   };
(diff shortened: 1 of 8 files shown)


tomasmoose:
Closing as this is invalid due to our updates in the latest satellite task.

01painadam:
Reopening.

01painadam:
Closing again, redundant.

PR #4201, opened 2021-03-30
add support for mixed env

tomasmoose:
Allows data to publish on staging+preprod at any time without deployment needed.


01painadam (approved):
works perfectly, please merge

PR #4205, opened 2021-04-07
fix/GFW-161

tomasmoose:
this pr fixes an issue where thresholds were not applied to widgets.

Cleaned up so we only care about `threshhold` now. We have to verify shaders where we use "thresh" [example](https://github.com/Vizzuality/gfw/blob/develop/providers/datasets-provider/config.js#L494). The previous behaviour was that `thresh` was the dynamic value, when user updates threshold in the UI and threshold was kept the original value at all times. Now I modiy threshold on runtime. If in the shaders we need to compare original `threshhold` i can store another key on fetch `originalThreshhold` or similar.

### DIFF
--- a/components/analysis/components/show-analysis/component.jsx
+++ b/components/analysis/components/show-analysis/component.jsx
@@ -57,7 +57,6 @@ class ShowAnalysis extends PureComponent {
     endDate,
     dateFormat,
     threshold,
-    thresh,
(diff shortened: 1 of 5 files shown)


01painadam (approved):
Tested, works as expected doesnt introduce any new bugs 👌

PR #4209, opened 2021-04-08
Remove second nav bar from chart

Bluesmile82:
## Overview

This is a little PR to remove the internal scrollbar on the embedded chart only on Trase. We removed the outer scroll but the internal one also needed to go

## Demo

Before:

image

After:

image

## Notes

This is not urgent. Just notify me when its deployed

## Testing

- It should be tested on Trase Profiles country pages after deploy


tomasmoose (approved):
Approved! will be deployed on Monday morning @Bluesmile82

PR #4219, opened 2021-04-28
Feat/historical weekly gfw 183

01painadam:
## Overview

Converts historic widget to weekly alerts.

Screen Shot 2021-04-28 at 10 16 25

Since daily alert fetches now times out at most Adm0 location we need to make lighter fetches. This PR creates a new widget - almost a carbon-copy of `fire-alerts-historical` - which instead make a much lighter fetch, aggregating alerts by week.


tomasmoose (approved):
Will push to pre-prod later today!

PR #4229, opened 2021-05-07
Feat/flux carbon widget

mluena:
## Overview

New Carbon Flux added to Dashboards/Climate

## Notes

Temporal images and colours until we have confirmation from the client

## Testing

http://localhost:3000/embed/widget/carbonFlux/country/BRA/?location=WyJjb3VudHJ5IiwiQlJBIl0%3D&widget=carbonFlux


components/widget/components/widgets-custom/carbon-flux/component.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+import Emissions from 'assets/icons/widgets/carbon-flux/emissions.svg';
+import Removals from 'assets/icons/widgets/carbon-flux/removals.svg';
```
01painadam, line 4: I think we still need to update to the svg's with labels?

components/widgets/climate/carbon-flux/index.js
```diff
@@ -0,0 +1,106 @@
...
+  POLITICAL_BOUNDARIES,
+  CARBON_FLUX,
+} from 'data/layers';
+
+import getWidgetProps from './selectors';
+
+export default {
+  widget: 'carbonFlux',
+  // TODO: global title
+  title: ' Forest-related greenhouse gas fluxes in {location}',
```
01painadam, line 22: Remove todo and whitespace, please :)

components/widgets/climate/carbon-flux/index.js
```diff
@@ -0,0 +1,106 @@
...
+  sentences: {
+    // TODO: global sentence
+    globalInitial:
+      'Between {startYear} and {endYear}, global forests emitted {totalEmissions}<b>/year</b>, and removed {totalRemovals}<b>/year</b>. This represents a net carbon flux of {totalFlux}<b>/year</b>.',
+    globalWithIndicator:
+      'Between {startYear} and {endYear}, global forests within {indicator} emitted {totalEmissions}<b>/year</b>, and removed {totalRemovals}<b>/year</b>. This represents a net carbon flux of {totalFlux}<b>/year</b>.',
+    initial:
+      'Between {startYear} and {endYear}, forests in {location} emitted {totalEmissions}<strong>tCO2e/year</strong>, and removed {totalRemovals}<strong>tCO2e/year</strong>. This represents a net carbon flux of {totalFlux}<strong>tCO2e/year</strong>.',
+    withIndicator:
+      'Between {startYear} and {endYear}, forests within {indicator}, {location} emitted {totalEmissions}<b>/year</b>, and removed {totalRemovals}<b>/year</b>. This represents a net carbon flux of {totalFlux}<b>/year</b>.',
```
01painadam, line 89: remove todo and make 'global' bold (as it's a location variable - these are always highlighted)

components/widgets/climate/carbon-flux/selectors.js
```diff
@@ -0,0 +1,273 @@
...
+
+    // calculate ticks array
+    const range = (start, stop, step) =>
+      Array.from(
+        { length: (stop - start) / step + 1 },
+        (_, i) => start + i * step
+      );
+    const ticks = Array.from(range(-tick, tick, 0.5)).map(
+      (t) => t * `1e+${maxValueDigits}`
+    );
```
01painadam, line 62: Nice solution! I guess since we're using the maxValueDigits to get the order of magnitude, we should do the same for the `xAxis.label.value` too?

i.e. replace G with M, k etc? 🤔  but that might be a pain. Alternatively we could use scientific notion in the tick labels (G,M,k etc) and make the axis label simple `tCO2e/year`

components/widgets/climate/carbon-flux/selectors.js
```diff
@@ -0,0 +1,273 @@
...
+    const { startYear, endYear } = settings;
+
+    const yearTotal = endYear - startYear + 1;
+    const { emissions, removals, flux } = data[0];
+
+    const params = {
+      indicator: indicator && indicator.label,
+      startYear,
+      endYear,
+      location: locationName === 'global' ? 'the world' : locationName,
```
01painadam, line 236: No need for this, right? Since the global sentences are hardcoded with global?

01painadam (changes requested):
Looking great - almost there!

components/widgets/climate/carbon-flux/selectors.js
mluena: totally right, I included it before including the global sentence (my left over)

components/widgets/climate/carbon-flux/selectors.js
```diff
@@ -0,0 +1,273 @@
...
+      },
+      referenceLine: { x: 0, stroke: '#606060', strokeWidth: 2 },
+      tooltip: [
+        {
+          label: 'Carbon flux (per year)',
+        },
+        {
+          key: 'removals',
+          label: 'Removals',
+          unitFormat: (value) => `${format('.3s')(value)}tCO\u2082e`,
```
tomasmoose, line 189: HI! looks great, i assume `u2082e` is superscript? Could you make a utils file called `unicode.js`

Then a constant something like: 

```
export const SUPER = `\u2082e`;
```

```
import unicode from 'utils/unicode';

unitFormat: (value) => `${format('.3s')(value)}tCO${unicode.SUPER}`
```

Because I know myself and I will screw one of these Unicode up in the future :)

01painadam:
OK, looks great! We just need to pull @tomasmoose 's api authentication changes and we're good to ship! Amazing work @mluena 👏

01painadam (approved):
🔥  Let's pull develop and we're golden!

components/widget/components/widgets-custom/carbon-flux/component.jsx
tomasmoose: @01painadam I can change these later if its needed

01painadam: it's resolved now, actually

tomasmoose:
Great work! will merge

PR #4237, opened 2021-05-18
fix global check

tomasmoose:
## Overview

This pr fixes an issue where we instead of reading global params, end up reading adm0, which in return gives us the wrong dataset `adm1` as target from `typeByGrouped`.


01painadam (approved):
Looks good! Can't see any issues.

PR #4241, opened 2021-05-24
add api key support for data api

tomasmoose:
## Overview

This pr adds two proxies to support the new way of dealing with API keys within GFW. 

## Testing, 

As all data requests now go through proxies, ensure data gets returned correctly through it. This affects any request that previously used `GFW_API`, `GFW_DATA_API` `GFW_STAGING_DATA_API` etc. 

When it comes to logic it affects **cashed analysis** + **OTF**.


tomasmoose:
Adding blocked will add proxy logic tomorrow morning.

tomasmoose:
> Adding blocked will add proxy logic tomorrow morning.

This is because with the current solution a tech-savvy person will be available to inspect + grab the API key. We will avoid this by running the API key through our own API.

01painadam (changes requested):
Tested the staging environment (`staging-data-api`):

dashboards :white_check_mark:
downloads :white_check_mark:
OTF :skull_and_crossbones: 

For OTF it seem that the geostore proxy doesnt work. The POST request (to return the geostore_id) is returning 404 in the proxy, but then the analysis is never resolving - getting stuck in an infinite loop.

tomasmoose:
Need to take a look at some 308's that are happening. I'm pretty sure it's related to trailing slash issues. + 400+ pages should not perform this check.

PR #4242, opened 2021-05-24
add global dashboard sentence

tomasmoose:
## Overview

This pr adds a global summary header by category. It acts as a widget but is not really a widget.

Screenshot 2021-05-24 at 12 53 16

## Testing

- Check each admin level on dashboard and make sure we are displaying the correct sentence
- As depending on category; We need to verify the correct sentence is shown for its category.
- Check that sentences gets feeded the correct properties for country + admin variables. 
- Ensure app does not crash in general
- Ensure widget looks good.


01painadam (approved):
Added area to sentence props as AoIs were not passing the label into `{area}`. Looks good

PR #4243, opened 2021-06-03
Feat/planet basemap discovery

tomasmoose:
This pr improves the visibility of satellite basemaps + adds a new way to scrub through imagery timeframes. 

Screenshot 2021-06-03 at 12 43 29


01painadam (approved):
Looking good!

PR #4248, opened 2021-06-10
fix carbon widget on map analysis summary

tomasmoose:
We needed the widget to be a bit more clever when it comes to simple mode, I updated the logic to support It for this widget.

+ as i had to remove the labels, just did not work in this tiny space, i added a simple legend that can be added for these analysis widgets on the map.


01painadam (approved):
Works great. Nice one!

PR #4257, opened 2021-06-28
fix geostore fetch

tomasmoose:


01painadam (approved):
Works on all dashboards and map 👌

PR #4266, opened 2021-07-01
fix widget title embed

tomasmoose:
## Overview

As previous fix broke dashboard summary sentence, i now send embed as a prop so we get access to the location within our embeds, without breaking dashboards.


01painadam (approved):
Nice one! Seeing the country location in the embed title and no nasty side effects 💪

PR #4275, opened 2021-07-12
disable analysis on dashboards

tomasmoose:


01painadam (changes requested):
Seems that clicking on the map opens an 'invisible' pop-up. You can still see the 'x' to close the interaction ands the white boarder.

Screen Shot 2021-07-12 at 14 07 03

tomasmoose:
> Seems that clicking on the map opens an 'invisible' pop-up. You can still see the 'x' to close the interaction ands the white boarder.
> 
> Screen Shot 2021-07-12 at 14 07 03

Good catch, have fixed this issue in the last commit

01painadam (approved):
Nice fix, works now!

PR #4276, opened 2021-07-13
Add flux widget

01painadam:
## Overview

Adds Carbon Flux Widget _on the fly_ (OTF) analysis for custom shapes on the map.

*Carbon Flux*
Screen Shot 2021-07-13 at 09 47 01

*Carbon Emissions*
Screen Shot 2021-07-13 at 11 24 51

## Testing

Open map, open `Climate` tab, select a layer in the  `CARBON FLUX` sub-section, create custom analysis by drawing an area.

## Notes

Analysis functions in `analysis-cached.js` using OTF are slightly different to Postgre table analyses since they use different fields and require `geostore_id` and `geostore_orging` query params.

Additionally, since they hit a different dataset route in the data-api, they require `staticStatement.table` to be passed in the `params`.

As such, these functions are labelled as `<function_name>OTF` to distinguish them from regular (postgres) table analysis functions.

### DIFF
--- a/components/analysis/components/show-analysis/styles.scss
+++ b/components/analysis/components/show-analysis/styles.scss
@@ -7,7 +7,6 @@
 
   .show-analysis-body {
     width: 100%;
-    height: 100%;
     min-height: rem(260px);
     padding: rem(20px);
   }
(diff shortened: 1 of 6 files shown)


01painadam:
All good now. Let's deploy on monday

01painadam:
@Tomas, apparently there is a browser related issue in Firefox - could you investigate

01painadam:
This has been approved. Merging.

PR #4277, opened 2021-07-15
add possibility to analyse layer shapes

tomasmoose:


01painadam (commented):
Works great, nice fix!

01painadam (approved):

01painadam:
Going to redeploy with ENV=production for testing with data api.

PR #4285, opened 2021-07-28
add fetch for glad latest within timeline

tomasmoose:


01painadam:
I guess we cant test until the layer is viewable?

tomasmoose:
Have made a small refactor on the timeline, now we send from+to params to simplify the date logic. Datepickers can use latest if latestUrl is present.

PR #4308, opened 2021-10-07
fix issue where area is to large but its not

tomasmoose:
An area can be approx 1BH, we where comparing with 1MH,


elpamart:
works for me :)

PR #4319, opened 2021-10-27
Add API keys to requests made to the GFW API

tomasmoose:
## IMPORTANT  

Before this PR is merged, we need to add `NEXT_PUBLIC_GFW_API_KEY` in Heroku's environment variables. 

-----

## Overview

This PR implements API keys in requests made to both [globalforestwatch.org](https://data-api.globalforestwatch.org/) and [resourcewatch.org](https://api.resourcewatch.org/) API's (particularly `api` and `data-api`) requests.

## In this PR  

- Added  a new `utils/proxies` file, where the proxy paths are set  
- Added and configured three `api/<path>` proxy pages  
  _which take requests and attach the `x-api-key` to the request headers_  
  - `api/gfw`  
  - `api/data`  
  - `api/metadata`  
- Updated the codebase to make use of these proxies  
- Minor cleanup & prettier/linting issue fixing  
- Updated the test GitHub workflow to set the `GFW_API_KEY` env variable  
  _necessary for the tests to run_

## Notes  

- This PR buillds on top of the work done before.   
- Some proxying was already implemented in `develop`/`master`, it was maintained   
  _eg: `planet-tiles`, `planet-mosaics`, etc. Remnants of proxy tests in `api/data-api` and `api/gfw-api`   
- Not all requests to these APIs require API keys, but it was easier to proxy them all, and it kept the codebase cleaner   
  _eg: running SQL queries on the `data-api` does require an API key, but getting information about the dataset did not_    
- Auth requests are being proxied, but there is a stray `/check-logged` that I wasn't able to track. These requests don't require an API key anyways.  
  _We can choose to "unproxy" or track it down later on_    
- There are a few files (notably html/css) that have been changed "unnecessarily".   
  These were done automatically to old files by VSCode's linter/formatter when merging `develop` into these PR.   
  I haven't quite figured out how to remove those changes, but they're in a separate commit.   

## Testing  

**IMPORTANT:**
This PR is running on `preproduction`. Unfortunately although the staging APIs are the ones with API key checking enabled, they are not updated (very old datasets). 

We can test that the API keys are attached locally against the staging APIs by disabling any failing widgets (mostly due to missing columns in the datasets), and check that authentication and the remaining widgets work. 

However, running against `preproduction` allows us to ensure the proxying works correctly, that merging the code would not break in production, and that all the requests that should be proxied are. 

**TESTING:**  

Verify that everything works as normal
    - Dashboard  
    - Authentication  
    - Widgets  
    - CSV data/metadata downloads  
    - Analysis (including geometries, draw on map)  
    - Geostore  
    - Viirs, modis, glad, integrated alerts  
    - Geostore  
    - RW dataset and metadata requests  
    - "My GFW" area, including custom geometries   

Verify that requests made to `api.resourcewatch.org` and `data-api.globalforestwatch.org` are being proxied through `api/gfw` and `api/data` respectively. 

_They look switched, I've choosen to kept the original `utils/request` utility names identical to prevent future confusion when working in the code, since these are what we are used to_  

Easiest way to test this is to open chrome's DevTools, Network Tab, check the "3rd-party requests", and verify that requests to those endpoints do not exist. The exception will be `check-auth` as mentioned in the notes above; which can be ignored as that endpoint doesn't seem to require an API key. 

## Tracking 

[FLAG-555 | Implement API keys](https://gfw.atlassian.net/browse/FLAG-555)


utils/proxies.js
```diff
@@ -0,0 +1,8 @@
+export const PROXIES = {
```
pedrogpimenta, line 1: @SARodrigues Great idea!

pedrogpimenta (approved):

PR #4330, opened 2021-12-13
add min absolute date for decoding integrated

elpamart:
## Overview

Update `getDayRange` function to include a `minDateAbsolut` param for decoding when this param is included in the`layerConfig.decodefuntion`. I've also updated the `layerConfig.decodefuntion` of the following layers to include this min value:

- Integrated alerts layer
- Glad l alerts layer
- Glad s alerts layer
- radd alerts layer

Finally, I've changed the Radd decode function for parsing this minDateAbsolut value from the layer config as min/max dayIndex instead of hardcoding the value directly in the code.

## Demo

## Notes

## Testing

### DIFF
--- a/components/map/utils.js
+++ b/components/map/utils.js
@@ -3,8 +3,10 @@ import { differenceInDays } from 'date-fns';
 import has from 'lodash/has';
 
 export const getDayRange = (params) => {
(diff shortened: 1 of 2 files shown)


tomasmoose (approved):
Great work!

PR #4350, opened 2022-02-14
Check geostoreId exists

pedrogpimenta:
## Overview

When loading the Areas Dashboard, the app would crash because it tries to call the API before having a valid `geostoreId` value. This PR will check that it exists before reaching the API.


elpamart:
Looks good to me!

PR #4396, opened 2022-06-08
[FLAG-350] Adjust login workflow to check cookie as well as local storage

pedrogpimenta:
This PR changes the way that the login is handled. 
It now adds a token validity date on the localStorage and logs out if it has expired.

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,7 @@
     "fast-redux": "^0.7.1",
     "file-saver": "^2.0.2",
     "final-form": "^4.18.7",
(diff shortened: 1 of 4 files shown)


pedrogpimenta:
For reference: this PR is being closed because it's old. The [corresponding ticket](https://gfw.atlassian.net/browse/FLAG-350) is Marked as Released as if we released this work. Maybe I forget and the issue was resolved some other way. Either way, this bug did not resurface for the past 6 months.

PR #4420, opened 2022-08-16
Change Tree Loss Fires Ranking name

pedrogpimenta:


pedrogpimenta:
I have no idea what this is or why I did this. Also there's a redirection added -- it shouldn't. Closing as weird and not documented.

PR #4438, opened 2022-09-21
[Implementation] [Research]Load Dashboard widget data on scroll (when widget enters the viewport)

SARodrigues:
## Overview

This PR makes it so that widgets do not load their data (and thus do not render charts) until they are in the viewport. 

This means that for a long page with multiple widgets, any widget initially visible in the viewport will load, and as the user scrolls the widgets will load their data as they become visible. 

## Testing

https://gfw-staging-pr-4438.herokuapp.com/dashboards

### Testing with Chrome dev tools

(Emptying cache & hard reloading between runs)

Screenshot 2022-09-23 at 12 18 28

## Tracking 

[FLAG-151](https://gfw.atlassian.net/browse/FLAG-151)


SARodrigues:
@pedrogpimenta I repurposed the investigation task (which already had the implementation). Only changes were that I merged `develop` into the PR. 

Conflicts found due to old age of this investigation were only on the `components/widget/index.js`, conflicts with the PR indented the charts on the net-change widget so special attention is required for that widget to make sure it still works properly.

PR #4446, opened 2022-10-05
col name changes tree cover density thres

AngelArcones:
## Overview

Column name changes:
Old: umd_tree_cover_density__threshold
New: umd_tree_cover_density_2000__threshold

## Testing

Global regions with most burned area widget (dataset = MODIS, change canopy density)


AngelArcones:
Not needed. Changes applied in a different PR.

PR #4472, opened 2022-11-24
[FLAG-499] Update GFW User data API to v2

pedrogpimenta:
## In this PR:

The `user` endpoint of the RW API has been updated to version 2. 
The structure changes a little bit, but I didn't want to change the behaviour of the Forms/Inputs, so I deconstructed the object when fetching and when sending it.

If you know how to clean up the construction of the object before sending to the API, I'd love to learn how! It's a bit... cumbersome.

## Testing

- Open your profile, and change stuff on the Update Profile modal window.


SARodrigues (approved):
Everything seems to work as expected! 

No idea how to clean it up better... I'd probably have followed a similar solution.

PR #4474, opened 2022-11-30
OTF analysis shows 0% decrease

pedrogpimenta:
## Overview

The extent area comes in a different format when querying pre-calculated area (as an array) or an OTF query (as a single value). This will will parse it one way or another depending on the type.

### DIFF
--- a/components/widgets/forest-change/tree-loss/selectors.js
+++ b/components/widgets/forest-change/tree-loss/selectors.js
@@ -24,10 +24,9 @@ const parseData = createSelector(
     if (!data || isEmpty(data) || !extentData) return null;
     const { startYear, endYear } = settings;
 
-    const extent =
-      (extentData && extentData.length && extentData[0].extent) ||
-      extentData ||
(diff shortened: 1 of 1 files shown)


SARodrigues (approved):
LGTM!

The only thing I can think of that can fail, that if `extentData` is an empty array, `extent` will be `[]` which may cause the widget to crash. 

I was doing some console testing to find an alternative and something like: 

```
Array.isArray(extentData) ? (extentData[0]?.extent || 0) : (!isNaN(extentData) ? extentData : 0)
```
would work for both possible scenarios (`extentData` being either a number or the array of objects you expect), but default to `0` if `extentData` is anything else (be it an object, empty array, string even), to prevent a crash.

**Edit:** but it is quite convoluted, at least visually.

pedrogpimenta:
@SARodrigues thanks! I guess it's better safe than sound. I edited your suggestion a little bit to not nest tertiary operations (linter doesn't let me :)

PR #4492, opened 2023-01-04
Fix www redirect only when it's globalforestwatch.org

pedrogpimenta:


server.js
```diff
@@ -33,7 +33,7 @@ function handleRedirectFor(urls, url, res) {
 function handleNonWwwToWwwRedirect(req, res) {
   try {
     const host = req.header('host');
-    if (!host.match(/^www\..*/i)) {
+    if (!host.match(/^www\..*/i) && host.match(/^globalforestwatch\..*/i)) {
```
SARodrigues, line 36: Only change I'd potentially make would be to move this down to line 56 

```
if (process.env.NEXT_PUBLIC_FEATURE_ENV === 'production') {
  handleNonWwwToWwwRedirect(req, res);
}
```

where we check whether we should apply the redirect, and add the why in the comments.

SARodrigues (approved):
The only concern I have is long term maintainability.

Do we really need to set `NODE_ENV` and `NEXT_PUBLIC_FEATURE_ENV` set to `production` on #4488, or is there another way? 

My concern is that by setting both to `production`, those environment variables can no longer be relied upon by a developer to check whether we're running the code in `staging`, `preproduction` or `production`. In a way that was the case already, and most likely we have both environment variables. 

I think this is fine (at least for now) albeit not ideal; we may want to rethink the setup later on. It'll be confusing for new developers.

server.js
pedrogpimenta: OK, it makes sense!

pedrogpimenta:
> The only concern I have is long term maintainability.
> 
> Do we really need to set `NODE_ENV` and `NEXT_PUBLIC_FEATURE_ENV` set to `production` on #4488, or is there another way?
> 
> My concern is that by setting both to `production`, those environment variables can no longer be relied upon by a developer to check whether we're running the code in `staging`, `preproduction` or `production`. In a way that was the case already, and most likely we have both environment variables.
> 
> I think this is fine (at least for now) albeit not ideal; we may want to rethink the setup later on. It'll be confusing for new developers.

Yes, I'm not sure why but `NODE_ENV` is always set to production everywhere, so, as you say, there's no way for a dev to know which is it. But I'd like to counter-argue that:

The possibility of having a test env act 100% like production feels very correct to me, it's a way to be almost 100% sure everything will work the same way, hence, I'd like to keep it that way!

PR #4505, opened 2023-02-08
[FLG-627] Grants and fellowships projets pagination

willian-viana:
## Overview

Wordpress by default only lets us get 100 posts at once. If G&F gets more Projects (it’s now at 94), they won’t load.

Instead of editing things so that the page fetches more project cards at a time, we’ve opted to paginate and load 21 posts at a time. We’ll follow the same styling that’s done on the GFW Blog (image below). Pagination should improve the loading speed of this page and probably help SEO-wise.

## Demo

grants-and-fellowships

## Notes
To meet the criteria we had to change the way we provide filters/search in this page.

- From now on, both search and filter are requesting WP Rest API for new items instead just order/omit items on front. 
- Countries filter logic was moved to SSR to avoid these computation on frontend side. 
- WP Rest API was changed and its code is on Pantheon. We should deploy it after that deploy.

## Testing

- Connect to WordPress locally and make sure you have all changes to get the advanced search feature.
- Visit /grants-and-fellowships/projects/

### DIFF
--- a/layouts/grants-and-fellowships/projects/component.jsx
+++ b/layouts/grants-and-fellowships/projects/component.jsx
@@ -11,8 +11,10 @@ import Dropdown from 'components/ui/dropdown';
 import TagsList from 'components/tags-list';
 import LoadMoreButton from 'components/load-more';
 
+import { getSGFProjects } from 'services/projects';
 import ProjectsModal from './projects-modal';
 import { getProjectsProps } from './selectors';
+
 
 import './styles.scss';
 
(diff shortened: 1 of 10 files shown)


SARodrigues (commented):
Pagination works great using the testing instructions (locally)! 
I found a couple strange behaviours that I'm not sure are related to the implementation. 

Clicking the search icon sometimes results in an error: 
https://user-images.githubusercontent.com/6273795/221592798-e7fa89ab-6160-4603-8955-0116185907b7.mp4
(In this recording it also looks like the message _"No projects for that search"_ shouldn't be there, since we are not searching; at least not at the same time as the loading spinner)

If the user searches, then changes countries, the search isn't applied. (not sure whether the search should be applied or cleared when changing countries)
https://user-images.githubusercontent.com/6273795/221593772-53857420-94c2-4099-9e70-cf791cef8d3c.mp4

When, with a country selected and a search, and there are no projects for that search, clearing the search box results in the _"Load more articles"_ button to appear, even though there are no results. 
https://user-images.githubusercontent.com/6273795/221594060-f80f23de-151e-4086-8447-182a98ae985f.mp4

pedrogpimenta (changes requested):
To add to what @SARodrigues said, if it helps: 

I think the search should reset when you change countries. It seems it does reset, but the field should visible reset as well.

Regarding the bug, it also happens when you select a country, do a search, then clear the search, not all projects appear again, only the ones that were there on the initial load. Ex: 
1. Load page
2. Select Brazil
3. Search for 2019
4. Clear Search
5. You only see 2 projects

Also: The "load more" button works and loads just some posts, so it's probably not using the load until 21 posts logic!

SARodrigues (approved):
LGTM! :)

layouts/grants-and-fellowships/projects/component.jsx
```diff
@@ -72,14 +112,46 @@ const GrantsProjectsSection = ({
...
   }, [country]);
 
+  useEffect(() => {
+    const { country: countrySelected } = query;
+    const isCountrySelectedOnFirstLoad = !!countrySelected;
+
+    if (isCountrySelectedOnFirstLoad) {
+      getMoreProjects({
+        search: countrySelected,
+        per_page: 100,
```
ltzenteno, line 130: wonder if it would be useful to move all these page values into a `const` given that they are being used several times in this file, i. e.

```
const ITEMS_PER_PAGE = Object.freeze({
  CUSTOM: 21,
  DEFAULT: 100,
});
```

and then in this line `per_page: ITEMS_PER_PAGE.DEFAULT`

ltzenteno (approved):

ltzenteno:
LGTM

pedrogpimenta (approved):
LGTM

PR #4516, opened 2023-03-02
Fix incorrect units in the Forestry employment donut chart

SARodrigues:
## Overview

This PR fixes the incorrect units in the Forestry employment donut chart. 
The number of people employed should was in `kha` when it should be in `k`. 

## Testing

TODO

## Tracking

[FLAG-534](https://gfw.atlassian.net/browse/FLAG-534)

### DIFF
--- a/components/widgets/land-use/forestry-employment/index.js
+++ b/components/widgets/land-use/forestry-employment/index.js
@@ -12,28 +12,29 @@ export default {
     {
       key: 'year',
       label: 'year',
-      options: [1990, 2000, 2005, 2010].map(y => ({ label: y, value: y })),
-      type: 'select'
-    }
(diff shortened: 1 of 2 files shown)


components/widgets/land-use/forestry-employment/index.js
```diff
@@ -12,28 +12,29 @@ export default {
...
   dataType: 'fao',
   metaKey: 'widget_forestry_employment',
   sortOrder: {
-    landUse: 3
+    landUse: 3,
   },
   settings: {
-    year: 2010
+    year: 2010,
+    unit: 'countsK',
```
SARodrigues, line 27: Note: This is the only relevant change in this file. The other changes were due VS Code auto-formatting.

PR #4517, opened 2023-03-02
Updated tropical tree cover layer | Toggle switch implementation

SARodrigues:
## IMPORTANT

https://github.com/Vizzuality/gfw/pull/4517/commits/bf4bfb1139d777b4da34c2cf8aa4e7a627a4cf0a needs to be removed before this PR is merged. It is only included at the moment to facilitate testing for both FE and Science. 

## Overview

This PR adds the possibility of using a toggle/switch in the map legend, in order to switch dataset layers. 

**In this PR:**
- Added the possibility to use a toggle switch to switch between a dataset's layers
  - It is enabled when the `isToggleLayer` property is set on a dataset  
  - Added a new `light-alternate` theme to the existing `<Switch />` component  
  - The `<Switch />` component can now also be displayed in a smaller size  
- Updated `gfw-components` to the latest version  
- Fixed an error _"state mutation detected between dispatches' error when switching layers"_
  _only is shown in development, made it impossible to switch layers_  
- Fixed an issue with `statementConfig` (layer statement) where the statement was built based on a dataset's default layer  
  _the correct configuration is to show the statement on a per layer basis_  
- Fixed an issue with the `<LayerStatement />` component, in which the `className` was not being applied for simple statements  
- Minor auto-linting fixes

## Screenshot

Screenshot 2023-03-03 at 15 28 13

## Test Environment

https://gfw-staging-pr-4517.herokuapp.com/

## Tracking

[FLAG-698](https://gfw.atlassian.net/browse/FLAG-698)

### DIFF
--- a/components/map/components/legend/component.jsx
+++ b/components/map/components/legend/component.jsx
@@ -228,7 +228,6 @@ const MapLegend = ({
                       layerGroup={lg}
                       name={name}
                       multi={isMultiSelectorLayer}
-                      toggle={isToggleLayer}
                       onChange={onChangeLayer}
                       {...selectorLayerConfig}
                     />
--- a/components/map/components/legend/component.jsx
+++ b/components/map/components/legend/component.jsx
@@ -168,6 +168,18 @@ const MapLegend = ({
                   />
                 )}
 
+                {isToggleLayer && selectorLayerConfig && (
+                  <LayerSelectorMenu
+                    className="layer-selector"
+                    layerGroup={lg}
+                    name={name}
+                    multi={isMultiSelectorLayer}
+                    toggle={isToggleLayer}
+                    onChange={onChangeLayer}
+                    {...selectorLayerConfig}
+                  />
+                )}
+
                 {activeLayer &&
                   paramsSelectorConfig &&
                   params &&
@@ -209,6 +221,7 @@ const MapLegend = ({
                     ) : null
                   )}
                 {(isSelectorLayer || isMultiSelectorLayer) &&
+                  !isToggleLayer &&
                   selectorLayerConfig && (
                     <LayerSelectorMenu
                       className="layer-selector"
--- a/components/map/components/legend/component.jsx
+++ b/components/map/components/legend/component.jsx
@@ -65,6 +65,7 @@ const MapLegend = ({
               isSelectorLayer,
               isMultiLayer,
               isMultiSelectorLayer,
+              isToggleLayer,
               selectorLayerConfig,
               color,
               metadata,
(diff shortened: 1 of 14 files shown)


services/datasets.js
```diff
@@ -15,7 +15,7 @@ const environmentString = () => {
 export const getDatasets = () =>
   rwRequest
     .get(
-      `/dataset?application=gfw&includes=metadata,vocabulary,layer&published=true&page[size]=9999&env=${environmentString()}${
+      `/dataset?application=gfw&includes=metadata,vocabulary,layer&published=true&page[size]=9999&env=production,preproduction,staging${
```
SARodrigues, line 18: This needs to be reverted before the PR is merged.

providers/datasets-provider/actions.js
```diff
@@ -20,7 +20,7 @@ const handleFeatureEnvLock = (env) => {
     return true;
   }
   if (currEnv === 'preproduction') {
-    return ['preproduction', 'production'].includes(env);
+    return ['preproduction', 'production', 'staging'].includes(env);
```
SARodrigues, line 23: This needs to be reverted before the PR is merged.

components/map/components/legend/index.js
```diff
@@ -90,13 +90,10 @@ class Legend extends PureComponent {
   onChangeLayer = (layerGroup, newLayerKey) => {
     const { setMapSettings, activeDatasets } = this.props;
     setMapSettings({
-      datasets: activeDatasets.map((l) => {
-        const dataset = l;
-        if (l.dataset === layerGroup.dataset) {
-          dataset.layers = [newLayerKey];
```
SARodrigues, line 96: This was the reason for the `state mutation detected between dispatches` error.

pedrogpimenta (changes requested):
- [x] Also, a reminder to remove the forced env variables from the request :)

components/map/components/legend/component.jsx
```diff
@@ -198,20 +212,23 @@ const MapLegend = ({
...
                   )}
                 {(isSelectorLayer || isMultiSelectorLayer) &&
+                  !isToggleLayer &&
                   selectorLayerConfig && (
                     <LayerSelectorMenu
                       className="layer-selector"
                       layerGroup={lg}
                       name={name}
                       multi={isMultiSelectorLayer}
+                      toggle={isToggleLayer}
```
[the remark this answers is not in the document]
SARodrigues: Good catch! You're right, it is not needed here; it'll be false inside the component. 
I'll remove it :)

PR #4555, opened 2023-05-05
FLAG-593: update limit startDate

wri7tno:
## Overview

Screenshot 2023-05-05 at 10 57 22 a m

Screenshot 2023-05-05 at 10 56 12 a m

## Demo

## Notes

## Testing

### DIFF
--- a/components/ui/datepicker/component.js
+++ b/components/ui/datepicker/component.js
@@ -2,6 +2,7 @@ import { useState, useRef, forwardRef } from 'react';
 import PropTypes from 'prop-types';
 import { Portal } from 'react-portal';
 import OutsideClickHandler from 'react-outside-click-handler';
+import moment from 'moment';
 
 import ReactDatePicker, {
   registerLocale,
@@ -50,7 +51,7 @@ const Datepicker = ({ lang, selected, minDate, maxDate, ...props }) => {
     </Portal>
   );
 
-  const selectedUTC = new Date(
+  let selectedUTC = new Date(
     selected.getTime() + selected.getTimezoneOffset() * 60000
   );
   const minDateUTC = new Date(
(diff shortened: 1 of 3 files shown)


components/ui/datepicker/component.js
```diff
@@ -60,6 +61,23 @@ const Datepicker = ({ lang, selected, minDate, maxDate, ...props }) => {
...
   );
 
+  const selectedMomentDate = moment(selectedUTC);
+  const minMomentDate = moment(minDateUTC);
+  const maxMomentDate = moment(maxDateUTC);
+
+  const diffFromStart = minMomentDate.diff(selectedMomentDate, 'days');
+  const diffFromEnd = maxMomentDate.diff(selectedMomentDate, 'days');
+
+  if (diffFromStart > 0) {
```
willian-viana, line 71: you don't need two ifs for this logic.

```suggestion
selectedUTC = minDateUTC;

  if (diffFromStart < 0) {
      selectedUTC = maxDateUTC;
  }
```

components/widgets/forest-change/integrated-deforestation-alerts/index.js
```diff
@@ -158,13 +160,23 @@ export default {
...
     const defaultEndDate = GLAD?.defaultEndDate;
-    const startDate = params?.startDate || defaultStartDate;
+    const selectedDate = params?.startDate || defaultStartDate;
     const endDate = params?.endDate || defaultEndDate;
 
     const isAoi = params?.locationType === 'aoi';
     const status = params?.status || 'unsaved';
     const isAnalysis = shouldQueryPrecomputedTables(params);
 
+    // overriding start date (FLAG-593)
```
willian-viana, line 170: I think for the same file you can use a method to avoid duplication.

`const setStartDateByAlertSystem = (alertSystem, params, selectedDate = null) => { ... }`

willian-viana (changes requested):

components/ui/datepicker/component.js
wri7tno: in this case `selectedUTC` already has a default value, which is 
```
new Date(
    selected.getTime() + selected.getTimezoneOffset() * 60000
  );
```

which is different than the if cases

hence we can't leave it with 1 if since we need to check both diffs

components/widgets/forest-change/integrated-deforestation-alerts/index.js
```diff
@@ -158,13 +175,20 @@ export default {
...
+    const { startDate, possibleStartDate } = setStartDateByAlertSystem(
```
willian-viana, line 186: cool 🚀

willian-viana (approved):
:shipit:

PR #4566, opened 2023-05-25
[FLAG-804] Update the Tree Cover Gain widget to show data for 2000 to 2012

willian-viana:
## Overview

Until we have an update from Data API with a bigger range we should use this solution to be more accurate in our info

## Demo

Screenshot 2023-05-25 at 11 57 08

## Testing

- Check the widget https://gfw.global/3J39LtR


wri7tno:
:shipit:

PR #4581, opened 2023-06-12
fix(metadata): change method replace to replaceAll

willian-viana:
## Overview

Using replaceAll to avoid errors when we have more than one occurrence of the same variable in metadata modal. Ex: [date]

### Before
Screenshot 2023-06-12 at 17 32 21

### After
Screenshot 2023-06-12 at 17 32 33

## Testing

- Open the map
- Select Tropical Tree Cover layer
- Open metadata info
- check the text


wri7tno (approved):
:shipit:

PR #4629, opened 2023-07-10
FLAG-852: display static image only in summary for IDN

wri7tno:
Display static image only in summary for IDN


willian-viana (approved):
:shipit:

PR #4650, opened 2023-07-30
FLAG-875: remove OTF query in getData to use precalculated q…

wri7tno:
Using precalculated tables instead of OTF queries in Carbon flux widget.


willian-viana (approved):
I left a suggestion but it's approved.

PR #4681, opened 2023-09-01
FLAG-900: Undefined year causing an error in the TCL due to Fires widget in protected and some custom area dashboards

SARodrigues:
## Overview

TODO

## Demo

TODO

## Notes

TODO

## Testing

TODO

PR #4685, opened 2023-09-05
FLAG-900: Undefined year causing an error in the TCL due to Fires widget in protected and some custom area dashboards

SARodrigues:
I accidentally merged #4681 into `develop` after review, but before QA. 
The accidental merged was promptly reverted by #4684.

This PR is to revert the revert after QA (and spin up a new test environment for the QA team).


SARodrigues:
Merging without review because it was already reviewed in #4681, but since it hadn't gone through QA I've reverted it.
As it has been now QA'd, I'm reverting the revert.

PR #4686, opened 2023-09-06
FLAG-866: fetch notifications and pass them as props to Header

wri7tno:
Fetching notifications for all pages. Notifications aren't visible yet, that is going to be addressed in a separate ticket.

## Overview

## Demo

## Notes

## Testing

### DIFF
--- a/services/__tests__/notifications.spec.js
+++ b/services/__tests__/notifications.spec.js
@@ -1,0 +1,60 @@
+import axios from 'axios';
+import { getPublishedNotifications } from '../notifications';
+import { NOTIFICATIONS_RESPONSE } from '../../fixtures/notifications';
+
+// eslint-disable-next-line no-undef
+jest.mock('axios');
+
+describe('getPublishedNotifications', () => {
+  describe('when API call is successful', () => {
+    it('should return a list of notifications', async () => {
+      const response = NOTIFICATIONS_RESPONSE;
+      const expectedResult = [
+        {
+          id: 16320,
+          dateCreated: '2023-09-05T22:44:03',
+          status: 'publish',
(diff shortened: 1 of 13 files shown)


services/__tests__/notifications.spec.js
```diff
@@ -0,0 +1,60 @@
+import axios from 'axios';
+import { getPublishedNotifications } from '../notifications';
+import { NOTIFICATIONS_RESPONSE } from '../../fixtures/notifications';
+
+// eslint-disable-next-line no-undef
+jest.mock('axios');
+
+describe('getPublishedNotifications', () => {
```
willian-viana, line 8: AWESOME! 🚀

services/__tests__/notifications.spec.js
```diff
@@ -0,0 +1,60 @@
...
+          status: 'publish',
+          title: 'Upcoming Webinar: Net Change in Tree Cover',
+          content:
+            '\n<p>Learn about the new net change in tree cover layer from WRI’s research team. After the presentation, participants will have the chance to ask questions. <a href="http://google.com">Sign up for the webinar here</a>.</p>\n',
+          icon: 'layers',
+          date: '2023-09-07',
+        },
+      ];
+
+      axios.get.mockResolvedValueOnce(response);
```
willian-viana, line 35: What happened when the result is a empty array? Do you think we need to create a test for this case?

willian-viana (approved):
:shipit:

services/__tests__/notifications.spec.js
wri7tno: not in this case since the empty array doesn't change anything, we are displaying a collection of items.

PR #4533, opened 2023-04-03
[FLAG-685] Planet obscures data layers

SARodrigues:
## Overview

This is a test branch to debug a possible dataset misconfiguration that causes layers to disappear in a very particular scenario. Notes on how to reproduce the original bug below. 

## Bug

- When Planet imagery is displayed with any other dataset, we’re finding that toggling PTW will cause that dataset to disappear. It will appear if that dataset is toggled on after PTW is toggled on or if the datasets are rearranged in the legend.
- When Planet is displayed, the PTW icons disappear when zooming in past zoom level 8. 
- Occurs across the following browsers: Edge, Edge in Private, Chrome, and Chrome incognito.

## Notes  

The layer configuration changes that seem to fix the bug can be found here: https://github.com/Vizzuality/gfw/pull/4533/commits/10065b93de5470ff0c3929f9a471f9e945f5db88 .
At the moment, this PR has mocked datasets in order to allow us to test the bug resolution. 

## Tracking

[FLAG-685 | Planet obscures data layers](https://gfw.atlassian.net/browse/FLAG-685)

## Testing

[...]

PR #4695, opened 2023-09-19
[FLAG-791] Issues turning the satellite imagery toggle on and off 

SARodrigues:
## Overview

This PR fixes an issue in sattelite imagery basemaps "stick". 

From what I've gathered the issue was due to the `key="basemap"` when adding the basemap layer; switching to `key={basemap.url}` also solves it.

I've chosen to merge the basemap back into the `allLayers` instead for a couple reasons: 
- I feel like it's clearer to pass an array of layers (basemap or not) to the LayerManager and let it handle them  
- Avoiding issues with keys in the future  
- It was done that way in the past. It was changed in an attempt to fix basemap issues in [#4533](https://github.com/wri/gfw/pull/4533), though it turned out that change only made the issue more frequent. The ultimate solution was found later on, with [#4633](https://github.com/wri/gfw/pull/4633), in which we finally figured out it was due to z-fighting issues when rendering decoded layers. 

## Demo

**Bug:**
https://github.com/wri/gfw/assets/6273795/98278616-5b1f-4f41-8fa2-f367f9b3e251

**Fix:**
https://github.com/wri/gfw/assets/6273795/d4071770-87e2-495e-b461-4f3d7baae422

## Notes

When rendering the PlanetAPI basemap/tiles, I did notice that they render a bit funny.
I've debugged this for a bit and it looks like a separate issue; the tiles come like that from the Planet API. I don't remember if it was like that before, though it is an unrelated issue. I've also tried to switch versions to past ones but they displayed the same way. 

## Tracking

[FLAG-791](https://gfw.atlassian.net/browse/FLAG-791)

### DIFF
--- a/components/map/components/layer-manager/component.jsx
+++ b/components/map/components/layer-manager/component.jsx
@@ -32,7 +32,7 @@ class LayerManagerComponent extends PureComponent {
           }
         : null;
 
-    const allLayers = layers.filter((l) => l);
+    const allLayers = [basemapLayer, ...layers].filter((l) => l);
 
     return (
       <LayerManager
@@ -69,9 +69,6 @@ class LayerManagerComponent extends PureComponent {
           },
         }}
       >
-        {basemapLayer && (
-          <Layer key="basemap" {...basemapLayer} {...basemapLayer?.config} />
-        )}
         {allLayers &&
           allLayers.map((l) => {
             const config = l.config ? l.config : l.layerConfig;


components/map/components/layer-manager/component.jsx
```diff
@@ -32,7 +32,7 @@ class LayerManagerComponent extends PureComponent {
           }
         : null;
 
-    const allLayers = layers.filter((l) => l);
+    const allLayers = [basemapLayer, ...layers].filter((l) => l);
```
wri7tno, line 35: so, if I understand correctly sending the basemapLayer alongside the array of layers doesn't affect the final result, right?

wri7tno (approved):

wri7tno:
LGTM

components/map/components/layer-manager/component.jsx
SARodrigues: You're right, it doesn't affect the result. What we're defining as basemaps are really just regular layers (in this case satellite imagery ones). 

The only thing we need to pay attention to is the [zIndex](https://github.com/wri/gfw/blob/develop/components/map/components/layer-manager/component.jsx#L31); it needs to be lower than any other layers'; the layer manager will order them by zIndex. zIndex for the layers is assigned [here](https://github.com/wri/gfw/blob/develop/components/map/selectors.js#L432-L452).

We'd usually add the basemap layer first (in the array is fine), because layers are usually stacked on top of each other depending of the order they are added in, but because we're using `zIndex`'s we don't have to worry too much about it.

PR #4697, opened 2023-09-27
[FLAG-818] Fix widget highlighting in the dashboard

SARodrigues:
## Overview

This PR fixes an issue in which, when on the Dashboard, clicking the "Share" button in a widget and copying the link, when using the link the widget won't be highlighted. 

It also fixes a tangential bug where when navigating to the Dashboard on a clean slate, and choosing the "Forest Change", the highlighted widget (by default) isn't the first one in the page. 

## Notes

This PR focuses solely on the highlighting widgets when using a direct link to the widget, as well as a tangential bug on the Forest Change tab. While investigating I found related issues, but it seems logical to focus on the scope of the task instead of making big changes.

Some things I've noticed and addressed that may be of relevance: 

**1. `pages/dashboards/[[...location]].js`**

I've noticed we had code for a hook with no dependencies to set a `ready` state variable to `true`. This was unused. After investigating, I've noticed that this is a common pattern throughout the app. Example in the [map page](https://github.com/wri/gfw/blob/develop/pages/map/%5B%5B...location%5D%5D.js#L177-L187) or the [widgets embed](https://github.com/wri/gfw/blob/develop/pages/embed/widget/%5Bwidget%5D/%5B%5B...location%5D%5D.js#L125-L131). 

I am not aware of when this was removed, but as the comments in the hook explain (and after investigation), this makes it so that, when we load the page (which in turns renders the `widgets` component), the store is fully updated. My understanding on this is a bit fuzzy, but because we set store values before, this will postpone `ready` being set to `true` by one cycle, delaying rendering of the Dashboard/Widgets by one cycle, ensuring that when they load the store is fully updated. 

I'm not sure this is the best solution as this is obscure, but it falls out of scope especially since multiple pages use this pattern. Should be looked into in the future, on a heavy refactor of the dashboard/map/widgets core components. 

I've restored the original pattern. 

**2. Dashboard/Widget categories (component)** 

This one was confusing, but when looking at the selectors for both of the Dashboard layout and Widget component, we seemed to have duplicate selectors for setting a category. 

I would imagine this is old leftover code; one was meant to store the Dashboard tab, the other to set the category to load widgets to. To make matters worse, not only the map seems to use a similar pattern, but selectors on both default to returning the location set in the query params which plays at odds to the usage of a `setWidgetsCategory` action we have and is used. 

Because the map uses the widget's component (and thus selectors as well), and because there is a provision to have a default `summary` category (which doesn't really make sense for the map/analysis IMHO), I chose to keep the `category` in the widget's store as the source of truth to _both_ dashboard tabs and widgets component. It was the easiest way to solve redundancy without a big refactor. 

I don't think this is a proper solution; the hopes here are to have _one_ source of truth, the easiest change that is least likely to break anything else, and that works. I've used a hook on the dashboard page to set the category when the query param changes, ensuring it gets synced correctly and eliminated any references to the dashboard category in order to prevent confusion as to which selector/action to use; there's only one now. 

**3. Dashboard page re-renders**  

While doing 2, I noticed that the original `useEffect` hook that sets the `ready` state missed an (empty) dependency array. This causes it to be ran _twice_ at load; adding an empty dependency array makes it only run _once_. I believe other sections of the website had this oversight, but I've skipped it to keep it to the scope of the task, and to prevent bigger code changes. It's something we need to check though.

**4. Widgets' default category**  

I've noticed that in [components/widgets/reducers.js](https://github.com/wri/gfw/blob/develop/components/widgets/reducers.js#L10), while it is set for the `category`'s initial state to be `summary`, this never really applied. Because either the category was being set from the query param (which could not exist) or in some parts of the code it was set to `null`, when loading the default category would be set to `undefined` or `null`. I've modified the reducer so that, if no category is set, it'll default to the one specified in the `initialState`, which in this case is `summary`. 

This (along with the other changes previously mentioned) also helps with [this comparison](https://github.com/wri/gfw/blob/develop/components/widgets/index.js#L94-L97), in which the `prevProps.category` would be undefined on page load, which would be setting the `activeWidget` to `null` thus removing the highlighting. 

**5. `forest-change` dashboard category issues**  

While testing widget highlighting, I noticed that when on the Dashboard, and choosing the _"Forest Change"_ tab, the net change widget would be highlighted (last one in the page) instead of defaulting for the first one. This was a mistake made by myself [here](https://github.com/wri/gfw/blob/develop/components/widgets/selectors.js#L679-L686) when we implemented subcategories for the Dashboard tabs. 

When visiting tabs without subcategories, the code would work correctly; the first widget from the array would be selected. However, for pages such as forest change in which we group widgets by subcategories (forest loss, forest gain, net forest change), the original code no longer applies. 

Because we do have a prop specifying whether the widgets should be grouped by subcategory or not, I've used the prop so that, if they are, we'll pick the first widget from the _first_ subcategory in the list. The subcategories are defined [here](https://github.com/wri/gfw/blob/develop/data/categories.json#L13-L27); changing their order in this file would change the ordering on the front end. 

## Testing

1. Share link
    - Visit the Dashboard
    - Choose _a_ tab that is not the "Summary" one (bug doesn't occur on that one)
    - On any widget, click the "Share" button, copy the link
    - On a new tab, paste the link
    - Verify that when the Dashboard loads, the widget is highlighted (scrolling is a separate issue)
3. Forest change tab
    - Visit the Dashboard
    - Click the "Forest Change" tab  
    - Ensure the first widget is the highlighted one

## Tracking

[FLAG-818](https://gfw.atlassian.net/browse/FLAG-818)


willian-viana (approved):
Nice work! :shipit:

PR #4698, opened 2023-09-27
[FLAG-895] Add API Key validation to switch between Prod and Staging environments

willian-viana:
## Overview

After some investigation on https://gfw.atlassian.net/browse/FLAG-860  we have found same API Key for both environments, Prod and Staging. 

So we need to change it to use the right key depending on which environment we are. 

## Demo

Screenshot 2023-09-27 at 14 03 42

## Notes

We need to add a new env property on Heroku

## Testing

- Open [the dashboard](https://gfw-staging-pr-4698.herokuapp.com/) to use any request you want
- Check in the logs if the api key is the right one

### DIFF
--- a/pages/api/data/[...params].js
+++ b/pages/api/data/[...params].js
@@ -3,8 +3,16 @@ import httpProxyMiddleware from 'next-http-proxy-middleware';
 import { GFW_DATA_API, GFW_STAGING_DATA_API } from 'utils/apis';
 import { PROXIES } from 'utils/proxies';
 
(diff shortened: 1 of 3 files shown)


wri7tno (approved):

wri7tno:
LGTM

willian-viana:
After some discussion, we decided to use the same property and just change the value of `NEXT_PUBLIC_GFW_API_KEY` directly on Heroku side.

PR #4707, opened 2023-10-16
[FLAG-943] Tropical Tree Cover analysis malfunctioning

willian-viana:
## Overview

Tree Cover Density widget doesn't work correctly on Analysis when threshold is added as parameter. 
This widget has its own threshold (>10 as default), but when a user selects Tropical Tree Cover layer and changes the threshold, a new parameter was added to the query, making the request fail.

We did a major refactor at the `getWHEREQuery` to improve readability and add some new validations.

## Demo

Screenshot 2023-10-17 at 15 45 47

Screenshot 2023-09-07 at 17 49 47

## Testing

- Open the map
- Select Tropical Tree Cover
- Change the threshold
- Try to see Tree Cover Density Widget (it should work normally)

### DIFF
--- a/services/__tests__/get-where-query.spec.js
+++ b/services/__tests__/get-where-query.spec.js
@@ -33,7 +33,9 @@ describe('getWHEREQuery', () => {
 
     const query = getWHEREQuery(params);
     const expected = "WHERE iso = 'PER' ";
+    const notExpected = 'AND umd_tree_cover_density_2000__threshold = 40 ';
 
     expect(query).toEqual(expected);
+    expect(query).not.toEqual(notExpected);
   });
 });
--- a/services/__tests__/get-where-query.spec.js
+++ b/services/__tests__/get-where-query.spec.js
@@ -1,36 +1,39 @@
 import { getWHEREQuery } from '../get-where-query';
 
 describe('getWHEREQuery', () => {
-  describe('Tree Cover Density query', () => {
-    let params;
+  it('should return an string with each parameter separated by AND', () => {
+    const params = {
+      type: 'country',
+      adm0: 'BRA',
+      locationType: 'country',
+      extentYear: 2000,
+      thresh: 30,
+      threshold: 30,
+      forestType: 'plantations',
+      dataset: 'annual',
+    };
+    const query = getWHEREQuery(params);
+    const expected =
+      "WHERE iso = 'BRA' AND umd_tree_cover_density_2000__threshold = 30 AND gfw_planted_forests__type IS NOT NULL ";
 
-    beforeEach(() => {
-      params = {
-        type: 'country',
-        adm0: 'PER',
-        pathname: '/map/[[...location]]',
-        locationType: 'country',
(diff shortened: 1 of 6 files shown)


services/analysis-cached.js
```diff
@@ -305,99 +313,123 @@ const getLocationSelect = ({
...
+
+  if (parameterKey === 'threshold') {
+    if (dataset === 'tropicalTreeCover') {
+      paramKey = 'wri_tropical_tree_cover__decile';
+    } else {
+      paramKey = 'umd_tree_cover_density_2000__threshold';
+    }
+  }
+
+  return paramKey;
```
wri7tno, line 337: cool, way better to move all that logic inside its own function.

services/analysis-cached.js
```diff
@@ -85,6 +85,14 @@ const SQL_QUERIES = {
...
 const ALLOWED_PARAMS = {
   annual: ['adm0', 'adm1', 'adm2', 'threshold', 'forestType', 'landCategory'],
+  treeCoverDensity: [
+    'adm0',
+    'adm1',
+    'adm2',
+    'threshold',
+    'forestType',
+    'landCategory',
+  ],
```
wri7tno, line 95: honest question, `treeCoverDensity` is a new param in the query...? or what is the purpose of adding this array here? trying to understand a bit more this logic.

services/analysis-cached.js
```diff
@@ -305,99 +313,123 @@ const getLocationSelect = ({
...
+   */
+  if (isTreeCoverDensity && paramKeys.includes('threshold')) {
+    paramKeys = paramKeys.filter((item) => item !== 'threshold');
+  }
+
+  paramKeys.forEach((parameter, index) => {
+    const isLastParameter = paramKeys.length - 1 === index;
+    const hasFilterOption = ['forestType', 'landCategory'].includes(parameter);
+    const value = hasFilterOption ? 1 : params[parameter];
+    const polynameMeta = allFilterOptions.find(
```
wri7tno, line 373: so, we still don't know what `polyName` refers to 🤷‍♂️

wri7tno (approved):

services/analysis-cached.js
willian-viana: I removed "polyname" variables but not all of them. Polynames for this method were filterOptions from `landCategory` or `forestType`. A weird name for a weird variable. We can remove all references of it later.

services/analysis-cached.js
willian-viana: Anytime we need a different value in the dataset argument, we need to create a different object on `AllowedParams` object. As I needed to validate treeCoverDensity dataset, I've created this object.

services/__tests__/get-where-query.spec.js
```diff
@@ -0,0 +1,41 @@
+import { getWHEREQuery } from '../get-where-query';
+
+describe('getWHEREQuery', () => {
```
wri7tno, line 3: 🎉

PR #4735, opened 2023-12-26
[FLAG-982] Weekly Fire Alerts wrong date comparison

willian-viana:
## Overview

When doing a year comparison for the[ weekly fire alerts widget](https://gfw.global/47ESukX), the selected year or the blue line shows a previous year. For example, 2022 was selected, but 2021 is shown instead. We must ensure that Dec 2022-Dec 2022 is displayed when 2022 is selected. I’m noticing the same pattern for all selected years. 

### Acceptance criteria:

- Check whether the data or just the label is incorrect
- Implement a fix so that the blue label shows a full year (in the example below, should be Dec 2021 - Dec 2022)

## Demo

Screenshot 2023-12-28 at 15 32 39

## Testing

- Open the [widget](https://gfw-staging-pr-4735.herokuapp.com/dashboards/country/BRA/?category=fires)
- Select the year to compare
- Validate which year appears at year label

### DIFF
--- a/components/widgets/utils/data.js
+++ b/components/widgets/utils/data.js
@@ -188,7 +188,7 @@ export const getStatsData = (data, latest) => {
   const lastYear = data.filter(
     (item) =>
       item.year === latestWeek.year() ||
-      (item.week >= latestWeek.isoWeek() && item.year >= latestWeek.year() - 1) // 52 weeks ago, including latestWeek
+      (item.week >= latestWeek.isoWeek() && item.year >= latestWeek.year()) // 52 weeks ago, including latestWeek
   );
 
   const parsedData = lastYear.map((d, i) => {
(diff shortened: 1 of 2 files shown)


components/widgets/utils/data.js
```diff
@@ -188,7 +188,7 @@ export const getStatsData = (data, latest) => {
   const lastYear = data.filter(
     (item) =>
       item.year === latestWeek.year() ||
-      (item.week >= latestWeek.isoWeek() && item.year >= latestWeek.year() - 1) // 52 weeks ago, including latestWeek
+      (item.week >= latestWeek.isoWeek() && item.year >= latestWeek.year()) // 52 weeks ago, including latestWeek
```
wri7tno, line 191: nice!, easy fix 😀

wri7tno (approved):

components/widgets/utils/data.js
willian-viana: yeah! Finally an easy ticket haha

PR #4736, opened 2024-01-09
fix(rw-api): remove x-api-key from RW API requests

willian-viana:
## Overview

RW API now validates Api Key. So for now, we are removing the x-api-key until we have the right value for that.

## Demo

Screenshot 2024-01-09 at 12 30 52
Screenshot 2024-01-09 at 12 31 01

### DIFF
--- a/utils/request.js
+++ b/utils/request.js
@@ -27,9 +27,7 @@ const RESOURCE_WATCH_API_URL = RESOURCE_WATCH_API;
 
 // At the moment, the API key is the same
 const GFW_API_KEY = process.env.NEXT_PUBLIC_GFW_API_KEY;
-const GFW_METADATA_API_KEY = GFW_API_KEY;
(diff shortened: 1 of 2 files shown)


wri7tno (approved):

gtempus (approved):
Thank you, @willian-viana. Glad this wasn't more work. 👍

utils/request.js
tanderegg, line 1: We need to remove the API key from any request that is going to the RW API, this includes: Line 45 (GFW_API_URL, which is the RW API), Line 72 (GFW_METADATA_API_URL, which is the RW API), Line 100 (GFW_API again, for auth), and the one already removed.  Only DATA_API_URL is the GFW Data API.  The others may not cause a problem currently, but I'm not positive so it is probably safer to remove them.

dmannarino (approved):

tanderegg (changes requested):

PR #4756, opened 2024-02-25
FLAG-1014: Change main subscribe form and PTW to use Ortto instead of Pardot

SARodrigues:
## Overview

**In this PR:**  
- Add/change/reorder fields in the `/subscribe` form  
- `interests` is now sent as an array of interests, instead of a string with comma separated positions  
- Updated the form handler link to the new one  
- Updated message displayed after the user successfully submits the form  
- Added the possibility to add descriptions to the select input  
- Removed an unused pardot extra field  
- The _"Places to Watch"_ link now points to this subscribe form  

## Feature relevant tickets

[FLAG-1014](https://gfw.atlassian.net/browse/FLAG-1014)

### DIFF
--- a/components/forms/newsletter/component.jsx
+++ b/components/forms/newsletter/component.jsx
@@ -151,10 +151,11 @@ class NewsletterForm extends PureComponent {
                     placeholder="Select a sector"
                     required
                   />
-                  <Checkbox
+                  <Select
                     name="interest"
                     label="I'm interested in (check all that apply)"
                     options={interestsOptions}
+                    multiple
                   />
                   <Error
                     valid={valid}
--- a/components/forms/newsletter/component.jsx
+++ b/components/forms/newsletter/component.jsx
@@ -15,6 +15,20 @@ import Error from 'components/forms/components/error';
 
 import { email as validateEmail } from 'components/forms/validations';
 
+const sectors = [
+  "Government",
+  "Donor Institution/Agency",
+  "Local NGO (National or Subnational)",
+  "International NGO",
+  "UN or International Organization",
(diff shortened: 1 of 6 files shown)


components/forms/newsletter/component.jsx
```diff
@@ -30,32 +55,35 @@ class NewsletterForm extends PureComponent {
     initialValues: PropTypes.object,
   };
 
-  saveNewsletterSubscription = (values) => {
+  saveNewsletterSubscription = async (values) => {
+    const ipAddress = await fetch('https://api.ipify.org/?format=json')
```
wri7tno, line 59: noob question, what is this url for?

wri7tno (approved):

wri7tno:
code LGTM, untested

components/forms/newsletter/component.jsx
SARodrigues: Not a noob question, this is a weird one. 

Task requirements list that we should send the _client_'s IP address as as a field when submitting the form.   
This is odd, but in order to do so (within the timeframe), when submitting the form, I'm pinging a third party service to get the ip address of the user (ran in the browser, so we are 100% sure it'll be the user's IP address). 

I went into specifics in the [JIRA task](https://gfw.atlassian.net/browse/FLAG-1014?focusedCommentId=24534) as to why this particular one, but it's basically because it's a free service, not rate limited, and the project is open source which fits. 

I do think a different solution should be procured in the future; and I'm not entirely sure sending the ip address is needed at all in the first place; Ortto should have this information already but this is a stepping stone.

PR #4813, opened 2024-06-21
[FLAG-973] Private saved areas are not accessible in a user's own My GFW account

willian-viana:
## Overview

A user shared their area with me: http://www.globalforestwatch.org/dashboards/aoi/6557f6b9695f90001af381ea (currently it’s set to public), but when a user chooses to change it to private, it’s no longer accessible: https://www.globalforestwatch.org/dashboards/aoi/6557f6b9695f90001af381ea/?lang=en in their own My GFW account.

## Notes

The issue was related to an specific cookie (gfw-token) that once expired, the user remained logged in because the same cookie is found in the localStorage. The token must be validated by cookies instead of local Storage, so I removed this validation to ensure that every time the cookie expires, a new login is required. 

Cookie was set to expires in 1 year.

### DIFF
--- a/components/modals/meta/actions.js
+++ b/components/modals/meta/actions.js
@@ -17,10 +17,11 @@ export const getModalMetaData = createThunkAction(
 
         getMetadata(metakey, metaType)
           .then((response) => {
+            console.log('XPTO', response);
             if (metaType === 'widget') {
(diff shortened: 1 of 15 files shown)


willian-viana:
PR with wrong commits.

PR #4836, opened 2024-08-01
[FLAG-923] Pasting a special character into search by location lead to an error occurred

SARodrigues:
## Overview

The FE makes use of [btoa()](https://developer.mozilla.org/en-US/docs/Web/API/Window/btoa) to Base64 encode parameters for the URL. However, [this method fails when strings include characters that take more than one byte](https://developer.mozilla.org/en-US/docs/Web/API/Window/btoa#unicode_strings), which is the case of _"Bình Thuận"_, the string that when searched caused the FE to crash. 

This is known as [the Unicode Problem](https://developer.mozilla.org/en-US/docs/Glossary/Base64#the_unicode_problem). 

The MDN Web Docs offer a solution, which passes by first encoding the string with [TextEncoder](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder) (and later using the [TextDecoder](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder) counterpart to reverse it), which is what this PR implements. 

I've created a separate util (`utils/url-param.js`) exclusive for this processing, with comments so that future developers are aware of this issue. I've chosen not to add these methods to `utils/url.js` as that file contains extra processing, including functions to remap legacy datasets, and I feel like the objective of these new very specific functions would not be clear. 

## Testing

 1. Visit https://gfw-staging-pr-4836.herokuapp.com/map
 2. On the sidebar, click "Search"  
 3. Fill in "Bình Thuận" in the search box  
 4. Verify that the application does not crash

## Tracking 

[FLAG-923](https://gfw.atlassian.net/browse/FLAG-923)

[FLAG-923]: https://gfw.atlassian.net/browse/FLAG-923?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/utils/url-param.js
+++ b/utils/url-param.js
@@ -1,0 +1,15 @@
+const encode = (param) => {
+  return btoa(JSON.stringify(param));
+};
+
+const decode = (param) => {
+  // we use Buffer as atob is node native to node for SSR
+  return JSON.parse(Buffer.from(param, 'base64').toString('binary'));
+};
+
+const urlParam = {
+  encode,
+  decode
+};
+
+export default urlParam;
--- a/utils/url-param.js
+++ b/utils/url-param.js
@@ -1,13 +1,26 @@
+// We encode url params into Base64 to add them to the URL by using window.btoa().
+// However, this method can fail if the string being encoded includes characters
+// outside of the Latin1 alphabet. Eg: "Bình Thuận".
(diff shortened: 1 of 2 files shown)


wri7tno (approved):
whoa, top notch Simao 👍

PR #4853, opened 2024-09-20
[FLAG-1193] Error in the query of TCL widget within some custom areas

willian-viana:
## Overview

The query of the widgethttps://gfw.global/4e4K0Xq, has undefined for the year, likely from an undefined JavaScript variable:

## Demo
Screenshot 2024-09-20 at 16 11 26


wri7tno (approved):
nice catch!

PR #4864, opened 2024-10-03
[FLAG-1098] Priority 2: Rename the “tree cover by type” widget and add “Plantations” intersection

willian-viana:
### Story | Acceptance criteria

As a user, I can view how much tree cover and other land cover exists in my area of interest for 3 tree cover datasets (UMD tree cover 2000 and 2010, tropical tree cover 2020). I can also see the area of tree plantations in my area of interest.

Given that:
- The user is viewing the [“tree cover by type” widget](https://gfw.global/47W8RKt) on the summary or land cover tabs.
- The user has selected any of the options in the “Tree cover dataset” selector in the settings of this widget (“Tree cover 2000”, “Tree cover 2010”, or “Tropical tree cover 2020”)

#### Widget title

- We will change the name of the widget to: Global: “Global tree cover”Country/subnational: “Tree cover in [area]
- Change categories shown on the widget
- The default categories will now be:
  - tree cover (hex: #506917)
  - other land cover (hex: #E7E5A4)

- When the user hovers over the donut chart, it will still show that category name and percent value, just as it does in the current setup. Here is an example from the current widget - natural forests won’t actually show up in the finished product:

#### Add “forest type” dropdown

In the “settings” option, there will be a new “forest type” dropdown

- When the “info”  button is clicked, the SDPT (“Tree plantations”) metadata will appear. 
- The only option in the “forest type” category will be “Plantations”, which will pull from the updated SDPT data.
- When "plantations" is selected as the forest type option, categories are:
  - tree cover in plantations (hex: #506917)
  - other tree cover (hex: #A0C746)
  - other land cover (hex: #E7E5A4)

**This matches the same setup you'll see in the https://gfw.global/3U181Wy . The query should match up too - whatever we find from https://gfw.atlassian.net/browse/FLAG-1135  will inform how we structure this query. We will update the query to match what we use in the [location of tree cover widget:](https://gfw.global/46ZpWRP)**

### DIFF
--- a/components/widgets/land-cover/tree-cover/index.js
+++ b/components/widgets/land-cover/tree-cover/index.js
@@ -24,9 +24,8 @@ import getWidgetProps from './selectors';
 export default {
   widget: 'treeCover',
   title: {
-    default: 'Tree Cover by type in {location}',
-    global: 'Global tree cover by type',
-    withPlantations: 'Forest cover by type in {location}',
+    default: 'Tree Cover in {location}',
+    global: 'Global tree cover',
   },
   alerts: [
     {
--- a/components/widgets/land-cover/tree-cover/index.js
+++ b/components/widgets/land-cover/tree-cover/index.js
@@ -108,7 +108,13 @@ export default {
     summary: 4,
     landCover: 1,
   },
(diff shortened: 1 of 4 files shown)


ltzenteno (approved):
code LGTM, gj

PR #4876, opened 2024-10-28
[FLAG-1121] soilOrganic widget to pull data from the precalculated tables instead of the carto ones

SARodrigues:
## Overview

This PR adds a new request to `analysis-cached` to get request the widget's data from the pre-calculated tables and adapts the widget to work with the new data. The original call `getSoilOrganicCarbon` found in `services/climate.js` was not removed as it is still used by the carbon stock widget. 

In order to keep the implementation/calculations (`selectors.js`), the implementation implements the same logic found in the `whrc-biomass` widget (albeit adapted), as it is a known good one with a similar display. 

## Testing  

Verify that the widget is pulling data from the precalculated tables, and not the carto query. 

## Screenshot

soil_organic

## Tracking
[FLAG-1121](https://gfw.atlassian.net/browse/FLAG-1121)

[FLAG-1121]: https://gfw.atlassian.net/browse/FLAG-1121?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/widgets/climate/soil-organic/selectors.js
+++ b/components/widgets/climate/soil-organic/selectors.js
@@ -1,147 +1,169 @@
 import { createSelector, createStructuredSelector } from 'reselect';
-import isEmpty from 'lodash/isEmpty';
+import uniqBy from 'lodash/uniqBy';
 import findIndex from 'lodash/findIndex';
-import { formatNumber } from 'utils/format';
+import isEmpty from 'lodash/isEmpty';
 import sortBy from 'lodash/sortBy';
+import { formatNumber } from 'utils/format';
 
 // get list data
 const getData = (state) => state.data;
-const getLocationName = (state) => state.locationLabel;
-const getAdm0 = (state) => state.adm0;
-const getLocationDict = (state) =>
-  state.adm0 ? state.locationData : state.childData;
-const getLocationObject = (state) => state.location;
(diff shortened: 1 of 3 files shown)


components/widgets/climate/soil-organic/selectors.js
```diff
@@ -1,147 +1,169 @@
...
 
-const getSortedData = createSelector(
-  [getData, getSettings],
-  (data, settings) => {
+export const getSortedData = createSelector(
+  [getData, getSettings, getAdm1, getAdm2],
+  (data, settings, adm1, adm2) => {
     if (isEmpty(data)) return null;
-    return sortBy(data, settings.variable).reverse();
+    // console.log('soil-organic', { data, settings })
```
wri7tno, line 25: nit: remove this console.log

wri7tno (approved):
code LGTM, I see it replaces carto 👍

wri7tno:
actually, maybe we can include the change to use the new function you created in the `carbonStock` widget

wri7tno:
disregard last comment, I think it is better to open a new PR to implement your new function in the carbon stock widget

PR #4909, opened 2025-01-05
FLAG-1210: replace carto query with precalculated table for soil organic widget

wri7tno:
## Overview

## Demo

## Notes

## Testing


services/analysis-cached.js
```diff
@@ -2747,6 +2749,55 @@ export const getBiomassStock = (params) => {
   }));
 };
 
+export const getSoilOrganicCarbon = (params) => {
```
willian-viana, line 2752: Is there a reason to remove this method from the climate.js file to add it here?

willian-viana (approved):
:shipit:

services/analysis-cached.js
wri7tno: that file was using carto, but I can move it to that file if we want to clean up this one

wri7tno: although it follows the very same logic all other functions in this file do, like using the sql queries at the top of the file and whatnot

PR #4919, opened 2025-01-16
[FLAG-1239]: Add GADM Query Parameter When GETing Areas

willian-viana:
## Overview

This card makes the GADM version explicit when a user’s Area is being requested from the Resource Watch (RW) Area microservice.

Making a distinction between different versions of administrative boundary IDs is critical to quickly adopting future boundary updates.

Currently, administrative Areas are referenced with GADM IDs. These are used when saving/loading Areas in Flagship. They’re also used in URL paths within Flagship for dashboards, embeds, and the like. Unfortunately, there is no guarantee from the GADM organization that IDs remain consistent across GADM versions.

For a made-up example, the same subregion in Mexico (or any other country) can have two sets of IDs depending on what GADM version is referenced:
GADM 3.6: MEX 8 14
GADM 4.1: MEX 9 12

This drastically affects Flagship's ability to keep a user’s areas consistent when changing to an updated GADM version.

Scope

This involves only adding two query parameters to a RW Area GET request.

Currently, requests for a user’s Areas look like this in Flagship: GET /v2/area

The updated implementation will add two query parameters: GET /v2/area?source[provider]=gadm&source[version]=3.6

Implementation Details

The RW Area microservice will ignore these parameters until later in the RW Area microservice GADM Epic.

However, the benefit is to get started on implementing the necessary interface updates and verifying integration with a low-risk approach.

It’s totally acceptable to hard-code these values to gadm and 3.6. Future cards will then make these values dynamic (to support 4.1).

Any other query parameters being sent for similar requests will still need to be sent. This is an addition only (in order to preserve backward compatibility).

Resources

[Miro sticky with comments](https://miro.com/app/board/uXjVLKsI-90=/?moveToWidget=3458764609708645038&cot=14)


wri7tno (approved):

services/areas.js
```diff
@@ -67,7 +68,9 @@ export const getAreas = () =>
 export const saveArea = (data) =>
   apiAuthRequest({
     method: data.id ? 'PATCH' : 'POST',
-    url: data.id ? `${REQUEST_URL}/${data.id}` : REQUEST_URL,
+    url: data.id
+      ? `${REQUEST_URL}/${data.id}${SOURCE}`
+      : `${REQUEST_URL}${SOURCE}`,
```
gtempus, line 73: These are `PATCH` and `POST` URLs. They don't need to have the new query params.
For these, a change in the body will be needed but, it's not part of the scope of this card.

services/areas.js
```diff
@@ -90,5 +93,5 @@ export const deleteArea = (id) => {
     action: 'User deletes aoi',
     label: id,
   });
-  return apiAuthRequest.delete(`${REQUEST_URL}/${id}`);
+  return apiAuthRequest.delete(`${REQUEST_URL}/${id}${SOURCE}`);
```
gtempus, line 96: Delete doesn't need the params, either.

gtempus (changes requested):
Looking good, @willian-viana.
Let's just focus on the `GET` requests for this PR.

services/areas.js
willian-viana: I just realized I replicated the source for all methods instead of GET only haha thank you for notice it!

gtempus (changes requested):
These changes are what I was thinking, @willian-viana. 💯 

Where are the tests? I do see that y'all have a few services tested.

Just as a warm-up, I added a basic areas.spec.js test for you to consider:

`services/__tests__/areas.spec.js`
```javascript
import { getArea } from '../areas';
import { apiAuthRequest } from "../../utils/request";

jest.mock('../../utils/request', () => ({
  apiAuthRequest: {
    get: jest.fn(),
  },
}));

describe('Areas Service', () => {
  describe('Getting a Single Area', () => {
    describe('The Request to the API', () => {
      it('should add source params for gadm 4.1', async () => {
        // arrange
        apiAuthRequest.get.mockResolvedValueOnce({data: {data: {attributes: {}}}});
        
        // act
        await getArea('abcdefg123456789');

        // assert
        expect(apiAuthRequest.get).toHaveBeenCalledWith(
          '/v2/area/abcdefg123456789?source[provider]=gadm&source[version]=4.1',
          {
            headers: {}
          }
        )
      });
    });
  });
});
```

Maybe add another for `getAll`? 😄

willian-viana:
> Your changes are what I was thinking, @willian-viana. 💯
> 
> Where are the tests? I do see that y'all have a few services tested.
> 
> Just as a warm-up, I added a basic areas.spec.js test for you to consider:
> 
> `services/__tests__/areas.spec.js`
> 
> ```js
> import { getArea } from '../areas';
> import { apiAuthRequest } from "../../utils/request";
> 
> jest.mock('../../utils/request', () => ({
>   apiAuthRequest: {
>     get: jest.fn(),
>   },
> }));
> 
> describe('Areas Service', () => {
>   describe('Getting a Single Area', () => {
>     describe('The Request to the API', () => {
>       it('should add source params for gadm 4.1', async () => {
>         // arrange
>         apiAuthRequest.get.mockResolvedValueOnce({data: {data: {attributes: {}}}});
>         
>         // act
>         await getArea('abcdef123456789');
> 
>         // assert
>         expect(apiAuthRequest.get).toHaveBeenCalledWith(
>           '/v2/area/abcdef123456789?source[provider]=gadm&source[version]=4.1',
>           {
>             headers: {}
>           }
>         )
>       });
>     });
>   });
> });
> ```
> 
> Maybe add another for `getAll`? 😄

Thank you so much, @gtempus! We've started the test culture years ago, unfortunately we've been struggling to keep it alive. I've added the tests for getAreas and getArea and I'll create tests for the other methods as well.

gtempus (approved):
Looks great, @willian-viana! 💥 
Thanks for adding the additional test and clearing the mock! 🙇

gtempus (approved):
Very nice, @willian-viana! 🤩 
Keeping up with the tests will actually make y'all go faster over time! ⏱️

PR #4926, opened 2025-02-04
[FLAG-1107] Add forest type 'plantations' to tree loss drivers widget

willian-viana:
## Overview

To be consistent with the SDPT layer being available as a “forest type” filter in the tree cover loss due to fires in [area] widget and forest-related greenhouse gas emissions in [area] widget, we would like to add it to the annual tree cover loss by dominant driver in [area] widget

#### Add “forest type” dropdown

We need to add a “forest type” dropdown in the settings for this widget. It will have two options:All tree cover (the default)PlantationsLuis and Will will see how easy it would be to make this change; it’s a new way of doing things. Please let Teresa know if this won’t work.

#### Add “Plantations” as intersection

In the “settings” button for this widget, there is a “Forest type” dropdown - we will need to add an option for “Plantations” just as we have for the tree cover loss due to fires in [area] widget and forest-related greenhouse gas emissions in [area] widget.

### DIFF
--- a/components/widgets/forest-change/tree-loss-drivers/index.js
+++ b/components/widgets/forest-change/tree-loss-drivers/index.js
@@ -49,6 +49,14 @@ export default {
   },
   settingsConfig: [
     {
+      key: 'forestType',
+      label: 'Forest Type',
+      whitelist: ['plantations'],
+      type: 'select',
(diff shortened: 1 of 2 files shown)


wri7tno (approved):
LGTM

PR #4930, opened 2025-02-12
[FLAG-1252] Use Data API Geostore Endpoints 

willian-viana:
## Overview

There are currently two geostore services at WRI. One is managed within [ResourceWatch](https://resource-watch.github.io/doc-api/reference.html#geostore) and the other within the [Data API](https://data-api.globalforestwatch.org/#tag/Geostore).

Data Lab’s strategic vision is to merge the two and have one single service managed solely by the Data API.

Because the GADM 4.1 administrative boundary geostores will be stored within the Data API, Flagship should update its geostore client from ResourceWatch to the Data API.

Scope

The Data API implements the endpoint interface that Flagship currently uses, so only a slight modification to the URL used will be necessary.

Implementation Details

Where Flagship makes calls like: https://api.resourcewatch.org/v2/geostore/...

It uses the Data API instead: https://data-api.globalforestwatch.org/geostore/...

The current gadm feature branch currently has a [Data API method](https://github.com/wri/gfw/blob/6283af5bce5b010b1614dc10ddf0dab48f1be9ad/services/geostore.js#L58) in geostore.js.  Since the ResourceWatch and Data API interface will be the same, the complicated dataset query logic will no longer be needed.  

 What About My Custom Geostores Managed within the ResourceWatch Geostore microservice?

The Data API transition will be gradual, starting with administrative boundaries. Data API Geostore service requests will proxy to ResourceWatch for any custom boundary request or other part of the existing interface. This will make our transition seamless for Flagship.


wri7tno (approved):

gtempus (approved):
Love the addition of the tests, @willian-viana ! 🔥 
I'm really glad this was a minimal change for y'all. 💪

PR #4939, opened 2025-02-25
[FLAG-1271] Add `simplify` query param when GETing admin geostores

willian-viana:
## Overview

This card makes the geojson simplification explicit when a geostore is requested from the Data API Area microservice.

Previously, Flagship passed a thresh query parameter when GETing administrative boundary geostores. The RW Geostore MS ignored the param because it wasn’t named correctly.

 It looked like it worked because the RW Geostore MS implemented its own default simplification values. I’ve documented [this on the miro board](https://miro.com/app/board/uXjVLKsI-90=/?moveToWidget=3458764605651383931&cot=14).

Going forward, we want to make sure we’re sending (and receiving) the simplified geostores that we’ve asked for and not rely on the RW Geostore MS defaults.

### Scope

This involves adding one query parameter to the Data API geostore GET request.

Currently, requests for a user’s Areas look like this in Flagship: GET /geostore/admin/:iso/:adm1/:adm2?source[provider]=gadm&source[version]=4.1

The updated implementation will add one query parameter: 

GET /geostore/admin/:iso/:adm1/:adm2?source[provider]=gadm&source[version]=4.1&simplify=<value>

This will work for GADM 3.6 requests, too

Implementation Details

[Here is the function](https://github.com/gfw-api/gfw-geostore-api/blob/a20d800e8083ff7f03e5b2d2a0840ad86fe65dce/app/src/services/cartoDBServiceV2.js#L62) used in the RW Geostore MS to calculate the default simplification value. 

The assumption is that this code can be copied over to the [services/geostore.js module](https://github.com/wri/gfw/blob/feat/gadm-geostore-service-FLAG-1252/services/geostore.js).

### DIFF
--- a/services/geostore.js
+++ b/services/geostore.js
@@ -24,21 +24,28 @@ const getWDPAGeostore = ({ id, token }) =>
     })
   );
 
+const parseSimplifyGeom = (iso, id1, id2) => {
+  const bigCountries = ['USA', 'RUS', 'CAN', 'CHN', 'BRA', 'IDN'];
+  const baseThresh = bigCountries.includes(iso) ? 0.1 : 0.005;
+  if (iso && !id1 && !id2) {
+    return baseThresh;
+  }
+  return id1 && !id2 ? baseThresh / 10 : baseThresh / 100;
+};
+
 /**
  * Fetch geostore from Data API
  * @param {string} url - geostore path e.g /geostore/admin/BRA/25/390
  * @param {string} token - Optional token for axios cancelToken
+ * @param {string} queryParams - Optional query parameters
  * @return {object} - An object with area id, geojson and bbox objects
  *
  */
-const fetchGeostore = ({ url, token }) =>
-  dataRequest
-    .get(
-      `https://data-api.globalforestwatch.org${url}?source[provider]=gadm&source[version]=3.6`,
-      {
-        cancelToken: token,
-      }
-    )
+const fetchGeostore = ({ url, token, queryParams }) => {
+  return dataRequest
+    .get(`https://data-api.globalforestwatch.org${url}?${queryParams}`, {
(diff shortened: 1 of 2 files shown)


wri7tno (approved):

services/geostore.js
```diff
@@ -48,28 +64,35 @@ const fetchGeostore = ({ url, token }) =>
...
 
 export const getGeostore = ({ type, adm0, adm1, adm2, token }) => {
   if (!type || !adm0) return null;
 
-  const adminURL = `/geostore/admin/${adm0}${adm1 ? `/${adm1}` : ''}${
-    adm2 ? `/${adm2}` : ''
-  }`;
+  const sourceProvider = 'source[provider]=gadm';
+  const sourceVersion = 'source[version]=3.6';
+  const threshold = `simplify=${setThreshold(adm0, adm1, adm2)}`;
```
gtempus, line 74: @willian-viana, it looks to me that a threshold was only ever applied to a `country` geostore. 
This PR seems to add it to every case (`country`, `use`, and `geostore`). 

I'm thinking this could result in unexpected geometries being returned from the Data API Geostore service.

I recommend applying the `simplify` query parameter to `country` geostore URLs **only**. Unless you see something I'm missing (and that's extremely possible). 😄

gtempus (changes requested):
I think the simplify query parameter is being applied to too many cases, @willian-viana.

services/geostore.js
willian-viana: You are right. I can't see a user drawing an area larger than a country like Brazil. So we don't need this parameter. I'll remove it asap

gtempus (approved):
Ship it! :shipit: 
Thanks, @willian-viana!

PR #4945, opened 2025-02-28
[FLAG-1261] Provide GADM Version when Creating/Updating an Area

willian-viana:
## Overview

The RW Area microservice and Flagship are aware of the concept of a provider and version for Areas that are administrative boundaries (GADM 3.6 or 4.1).

Therefore, when areas are created or updated using the Areas microservice, it’s necessary to provide the provider and version as part of the Area payload.

Implementation Details

[Areas Service where Areas are saved/updated](https://github.com/wri/gfw/blob/2d31f12257cc0783b85938bd662ad49acf2f3f52/services/areas.js#L68) in Flagship

Area payloads should have a source attribute in either its iso or admin object (or both) when saving/updating an Area representing an administrative boundary.

This schema is the same shape as what is currently returned in an Area representation from the MS:

```
"iso": {
  "source": {                   <= If the new area has `iso` values, add a `source` property
    "provider": "gadm",
    "version": "4.1"
  },
  "country": "HND",
  "region": "8",
  "subregion: "4"
},
"admin": {                      <= If the new area has `admin` values, add a `source` property
  "source": {
    "provider": "gadm",
    "version": "4.1"
  },
  "adm0": "HND",
  "adm1": 8,
  "adm2": 4
},
```

The Miro board has an [example of an Area Representation](https://miro.com/app/board/uXjVLKsI-90=/?moveToWidget=3458764609164177591&cot=14) with the source properties.

### DIFF
--- a/components/forms/area-of-interest/actions.js
+++ b/components/forms/area-of-interest/actions.js
@@ -15,124 +15,130 @@ import { getWidgetsData } from 'components/widgets/actions';
 export const saveAreaOfInterest = createThunkAction(
   'saveAreaOfInterest',
   ({
-    id,
-    name,
-    tags,
-    email,
-    webhookUrl,
-    language,
-    alerts,
-    admin,
-    wdpaid,
-    use,
-    application,
-    viewAfterSave,
-    publicArea,
-    deforestationAlertsType,
-    geostore: geostoreId,
-  }) => (dispatch, getState) => {
-    const { location, geostore } = getState();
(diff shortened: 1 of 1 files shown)


gtempus (approved):
Yes, this is what the Areas MS is expecting. 💯 

You're going to update this in the GADM 4.1 feature branch, too, right?

wri7tno (commented):
lgtm

PR #4956, opened 2025-03-07
[FLAG-1274] Update Areas Service to get both 3.6 and 4.1 User Areas

willian-viana:
### RW Areas MS Behavior

The RW Areas MS allows clients to filter administrative areas with a source[provider] and source[version] filter.

Regardless of the filter, all custom areas are always returned

For backward compatibility, when the filter isn’t used OR it’s used with the values gadm and 3.6, administrative boundary areas (those with iso or admin values) with a 3.6 version will also be returned. Any user’s administrative boundary that is only 4.1 will NOT be returned. This is so legacy clients do not try to interpret the 4.1 GADM IDs as 3.6 

Going forward, when the filter is used for values gadm and 4.1, administrative boundary areas with a 4.1 version will be returned. Any user’s administrative boundary area that is only 3.6 (areas that could not be automatically updated to 4.1) are NOT returned.

### Flagship Changes

Therefore, when Flagship wishes to update to GADM 4.1, it’s necessary to make two calls to the RW Areas MS. One with a filter of 3.6 and another with a filter of 4.1 and then create a union of the two sets to pass to the Redux store.

Here is a [basic diagram in the GADM miro board](https://miro.com/app/board/uXjVLKsI-90=/?moveToWidget=3458764618749660950&cot=14) to explain the operation.

### Scope

This should only affect the services/areas.js [module](https://github.com/wri/gfw/blob/d4d62a0e24c3af11b5a4d85441057385775b984a/services/areas.js). Specifically, getting Areas that are later loaded into Redux.

### DIFF
--- a/services/areas.js
+++ b/services/areas.js
@@ -2,11 +2,34 @@ import { apiAuthRequest } from 'utils/request';
 import { trackEvent } from 'utils/analytics';
 
 const REQUEST_URL = '/v2/area';
-const SOURCE = `?source[provider]=gadm&source[version]=3.6`;
+
+const parseArea = (area) => {
+  const admin = area.attributes && area.attributes.admin;
+  const iso = area.attributes && area.attributes.iso;
+
+  return {
+    id: area.id,
+    ...area.attributes,
+    ...{
+      ...area.attributes,
+      use: {},
+    },
+    ...(iso &&
+      iso.country && {
+      admin: {
+        adm0: iso.country,
+        adm1: iso.region,
+        adm2: iso.subregion,
+        ...admin,
+      },
+    }),
+    userArea: true,
+  };
+};
 
 export const getArea = (id, userToken = null) =>
   apiAuthRequest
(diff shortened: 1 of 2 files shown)


services/areas.js
```diff
@@ -15,55 +38,37 @@ export const getArea = (id, userToken = null) =>
...
+    })
+
+  return Promise.all([gadm36, gadm41]).then(([areas36, areas41]) => {
+    // logic to return only unique values, preferring 4.1 over 3.6
+    const areasMap = new Map();
+
+    areas36.map(area => areasMap.set(area.id, area));
+    areas41.map(area => areasMap.set(area.id, area));
+
+    return Array.from(areasMap.values());
```
gtempus, line 69: This looks great, @willian-viana! 💯 
Nice and clean.

gtempus (approved):
Nice implementation, @willian-viana !
:shipit:

PR #4958, opened 2025-03-07
chore(gadm): comment gadm 4.1 logic inside getAreas

willian-viana:
## Overview

Since we're still on 3.6 in production. We need to remove the second 4.1 request to the RW Areas MS.

#4957

### DIFF
--- a/services/__tests__/areas.spec.js
+++ b/services/__tests__/areas.spec.js
@@ -115,14 +115,16 @@ describe('Areas Service', () => {
               adm0: 'BRA',
               source: {
                 provider: 'gadm',
-                version: '4.1',
+                // version: '4.1',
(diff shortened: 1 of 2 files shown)


gtempus (approved):
I really like how you handled this, @willian-viana. 💪 
Going to production with the code like this will let us exercise a lot of your original implementation without the 4.1 side effect.

We can't have Flagship make 4.1 area queries because there are actually 4.1 areas in the Areas microservice. So, if Flagship starts asking for them in the areas service but isn't prepared to use 4.1 logic in other places consistently, it will cause confusion.

We should introduce the 4.1 area request in a GADM 4.1 feature branch. Just not into `main` right now.

PR #4973, opened 2025-04-08
[FLAG-1174] Remove repeat words in integrated alerts widget

willian-viana:
## Overview

When any individual alert system (GLAD-S2, GLAD-L, RADD) is selected in the [Integrated Deforestation Alerts in Brazil widget](https://gfw.global/3HVq7nm), an extra “high confidence alerts” appears after the widget sentence’s period. This underlined area that produces a modal when a user hovers it should replace the bolded high confidence alerts” text before it. We want users to be able to hover over “high confidence alerts” and see this explanation (“Alerts are classified as high confidence when…”).

## Demo

Screenshot 2025-04-08 at 13 38 36


wri7tno (approved):
:shipit:

PR #4994, opened 2025-05-07
[EPIC] TCL

willian-viana:
## Overview

Epic branch to centralize all 2025 TCL changes.

### DIFF
--- a/layouts/dashboards/components/header/selectors.js
+++ b/layouts/dashboards/components/header/selectors.js
@@ -12,7 +12,7 @@ import {
 } from 'providers/areas-provider/selectors';
 
 const isServer = typeof window === 'undefined';
-const DOWNLOAD_VERSION = '2023';
+const DOWNLOAD_VERSION = '2024';
 
 // get list data
 export const selectLocation = (state) =>
@@ -128,7 +128,7 @@ export const getDownloadLink = createSelector(
       }.xlsx`;
     }
 
-    return `https://gfw2-data.s3.amazonaws.com/country-pages/country_stats/download/gfw_2023_statistics_summary_v30102024.xlsx`;
+    return 'https://gfw2-data.s3.amazonaws.com/country-pages/country_stats/download/2024/global_05212025.xlsx';
   }
 );
 


components/charts/components/pie-chart-legend/component.jsx
```diff
@@ -50,12 +50,22 @@ class PieChartLegend extends PureComponent {
...
+                )}
+              >
                 {categoryItems.map((item, index) => {
                   const value = `${formatNumber({
                     num: item[config.key],
                     unit: item.unit ? item.unit : config.unit,
                     spaceUnit: item.unit !== '%' && config.unit !== 'countsK',
+                    ...(item.suffix && {
+                      suffix: item.suffix,
+                    }),
```
wri7tno, line 68: I added this suffix field so it can support examples like this (with the per year string):
```
14.5 MtCO2e per year
```

components/charts/pie-chart/component.jsx
```diff
@@ -24,7 +25,7 @@ class CustomPieChart extends PureComponent {
 
     return (
       <div className={cx('c-pie-chart', className)} style={chart?.style}>
-        <ResponsiveContainer width="99%" height={maxSize}>
+        <ResponsiveContainer height={230}>
```
wri7tno, line 28: I set it to 230 for now (so the size is consistent for both tcl drivers and emission drivers widgets, I was debugging and for some reason the tcl drivers widget had the correct size "230" but emissions had "140" somewhere in the code there was doing something wrong

services/datamart.js
```diff
@@ -52,6 +52,11 @@ const getDataByParams = async ({ dataset, aoi, canopy }) => {
         subregion: aoi.subregion,
       }),
     },
+    dataset_version: { // TODO: these hardcoded params will be removed soon since they'll be default after the TCL release, ask Daniel Mannarino
```
wri7tno, line 55: we will need to get rid of these params once Daniel Mannarino confirms that they are used by default on their side ☝️

services/datamart.js
```diff
@@ -108,6 +113,11 @@ const createRequestByParams = async ({
...
```
wri7tno, line 116: same here

wri7tno (approved):

PR #4995, opened 2025-05-09
FLAG-1325: remove custom IDN image and caveats

wri7tno:
## Overview

## Demo

## Notes

## Testing

### DIFF
--- a/components/widgets/forest-change/tree-loss-primary/index.js
+++ b/components/widgets/forest-change/tree-loss-primary/index.js
@@ -206,12 +206,7 @@ export default {
    */
   // eslint-disable-next-line no-unused-vars
   isPlaceholderImage: ({ location, category }) => {
-    return (
-      location.type === 'country' &&
-      location.adm0 === 'IDN' &&
-      !location.pathname.includes('embed') &&
-      (category === 'summary' || !category)
-    );
+    return false;
   },
   getWidgetProps,
 };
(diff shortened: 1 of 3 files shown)


components/widgets/forest-change/tree-loss-primary/index.js
```diff
@@ -206,12 +206,7 @@ export default {
    */
   // eslint-disable-next-line no-unused-vars
   isPlaceholderImage: ({ location, category }) => {
```
willian-viana, line 208: we are validating whether isPlaceholderImage is undefined so I believe it would be better to simply remove it instead of returning false.

willian-viana (changes requested):

components/widgets/forest-change/tree-loss-primary/index.js
wri7tno: I left it like that to have visibility on the attribute for future references like, if we need to send an image again sometime in the future, but I can remove the attribute as well.

wri7tno: only that we will lose visibility on that attribute since, you know... JS 🤷‍♂️

willian-viana: it makes sense! let's keep it there

willian-viana (approved):
:shipit:

PR #5230, opened 2026-07-10
PZB-905: fix Multiple Protected Area Analytics Issues on Flagship Map

guilherme-n:
## Overview

3 fixes for map:
1- TCL Graph Starting in 2002 Instead of 2001 TCL graphs for PA analytics start in 2002 rather than 2001.
2- Tree Cover Gain Showing 0.0% TC gain appears incorrect for PA analytics, even if TC in the 20-year dataset was 100%, the gain should still show more than 0.0% when looking at the other stats.
3- TCL by Dominant Driver Not Working TCL by dominant driver does not appear to be working for PA analytics.

## Demo

Screenshot 2026-07-10 at 20 14 13

__________

Screenshot 2026-07-10 at 20 15 15

__________

Screenshot 2026-07-10 at 20 16 08


01painadam (approved):
- TCL year looks good for PA
- Gain text is correct
- TCL Driver text is correct though it looks like theres a BE analytics issue


https://gfw-staging-pr-5230.herokuapp.com/api/data/dataset/geostore__tcl__change/v20260407/query/?sql=SELECT *
FROM data
WHERE geostore__id = '4fcf1120447801c912db6864c53a1e4b'

returns empty [] with success when though I see data on the map

PR #5231, opened 2026-07-20
PZB-1109: Update text from global forest watch to global nature watch & gfw to gnw

guilherme-n:
## Overview

Update text from global forest watch to global nature watch & gfw to gnw


gtempus (approved):
Thanks, @guilherme-n ! 🙇

